PR #6, opened 2018-02-05
Addworker cmd

cice:
Hey, so this is more proof of concept and pretty much my first Go code at all. 
So, room for improvement for sure, but maybe we can make sth of it.
Cheers!
Marian

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -22,7 +22,7 @@ import (
 var clusterCmd = &cobra.Command{
 	Use:   "cluster",
 	Short: "view and manage kubernetes clusters",
-	Long:  `This command bundles several sub-commands to handle with kubernetes clusters, running on Hetzner Cloud.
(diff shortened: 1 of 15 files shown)


xetys:
Hey! Thanks a lot for your contribution. The code looks very good. 

I tested the code and noticed that your version doesn't run the node provisioning in parallel. When you create a new cluster, it directly starts to provision the last worker node. Consequently, it fails to provision an initial cluster.

And a little convention stuff: 

* could you please use 'AddWorker" instead of "Addworker"
* would you kindly add a cobra description? You can remove the long description :smile: 

As soon the issues are solved, I'll merge this.

cice:
Hey, thanks alot for the feedback. I implemented the requested changes, i think it works properly now.

xetys:
there are some little issues, but I will fix them on my own. LGTM and thanks for your contribution!

PR #7, opened 2018-02-16
Added support for cloud-init

hg8496:
Its my first go code. So please examine closely.


cmd/cloudInitSet.go
```diff
@@ -0,0 +1,49 @@
...
+// limitations under the License.
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"errors"
+)
+
+// clusterAddWorkerCmd represents the clusterAddWorker command
```
xetys: wrong description

xetys:
Hey and thank you for your contribution!

I like the idea of using the cloud-init feature, but here are my 2 thoughts on that:
1. implemented this way the file affects all future servers, unaware of the cluster. I think putting it under `cluster cloud-init` would give more control.
2. how complex it would be to use the shell-script `install-docker-kubeadm.sh` to that cloud-init and make further stuff injectible to that configuration?

Cheers

hg8496:
Hey,

your welcome.

My first solution was as you described in (1.) but I was unsure where to save the information. Probably it needs to be a parameter to `cluster create` since we need a `Cluster` to store the information.
Another approach could be a separate configure structure for cloud-init. But I think that's too much code duplication. 
Any thoughts?

As of (2.) I think that should be possible. As far as i know there is multi-part support in cloud-init.
I am not sure what the advantage would be. One would still need the ssh connection to connect master and client. With the direct execution you have direct feedback from your script.

Greetings

hg8496:
I just gave the `cluster create` approach a try. I think it turned out the best solution.
Less code. Parameter the place it is needed.

hg8496:
Anything waiting in me?

xetys:
Was just knocked out by sickness, sry for delay

PR #9, opened 2018-02-18
[MOD] deleting useless dependencies to make it compile on newer go ve…

pushthat:
Hey,

The commit message is pretty explicit.

Regard,


xetys:
yes, definitely!

Issue #14, opened 2018-02-25
HA support


xetys:
In the last 10 hours, I successfully built a true HA cluster manually. 

To do this, these steps I noted down are needed to provision a full HA kubernetes:

* (Optional) deploy WireGuard for secured node networking
* Bootstrap etcd cluster on the master nodes. At least three should be used
* init first master using that cluster and extended SAN names
* distribute master PKI to other master nodes
* deploy xetys/k8s-master-lb to all nodes
* reconfigure api server to have `- -apiserver-count 3`
* reconfigure kube-proxy to use the master-lb
* reconfigure kubelet.conf, controller-manager.conf and scheduler.conf to use the master-lb
* restart all kubelets
* join worker nodes using balanced

in particular this setup was tested with heavy network disruptions and kubernetes was still able to operate and is stated as highly available.

A different approach would be using a Floating IP and assign it by keepalived. That would cost extra $

PR #22, opened 2018-02-27
implement HA support

xetys:
related to #14


cmd/clusterCreate.go
```diff
@@ -37,8 +37,11 @@ to quickly create a Cobra application.`,
 	PreRunE: validateClusterCreateFlags,
 	Run: func(cmd *cobra.Command, args []string) {
 
-		nodeCount, _ := cmd.Flags().GetInt("nodes")
-		workerCount := nodeCount - 1
+		workerCount, _ := cmd.Flags().GetInt("worker-count")
+		masterCount, _ := cmd.Flags().GetInt("master-count")
+		etcdCount, _ := cmd.Flags().GetInt("etcd-count")
+		haEnabled, _ := cmd.Flags().GetBool("ha-enabled")
+		isolatedEtcd, _ := cmd.Flags().GetBool("iso-etcd")
```
pierreozoux: isolated-etcd ?

xetys: I thought it is unhandy to type a word of this length...thing a full-length is better?

cmd/clusterCreate.go
```diff
@@ -171,26 +225,49 @@ func validateClusterCreateFlags(cmd *cobra.Command, args []string) error {
...
-
-	// Cobra supports Persistent Flags which will work for this command
-	// and all subcommands, e.g.:
-	// clusterCreateCmd.PersistentFlags().String("foo", "", "A help for foo")
-
-	// Cobra supports local flags which will only run when this command
-	// is called directly, e.g.:
 	clusterCreateCmd.Flags().String("name", "", "Name of the cluster")
-	clusterCreateCmd.Flags().StringP("ssh-key", "k","", "Name of the SSH key used for provisioning")
+	clusterCreateCmd.Flags().StringP("ssh-key", "k", "", "Name of the SSH key used for provisioning")
```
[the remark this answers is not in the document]
xetys: this was more unintended, than `ctrl+alt+t` in IDEA

pierreozoux:
This looks amazing :) If you need help to test? Happy to help!

cmd/clusterCreate.go
pierreozoux: I usually see projects like etcd or kubernetes using short version of some words, and later on, switch back to full length.
I think it is better to avoid confusion. When I see short version, sometimes, I always need to go and check the documentation to see to what it refers.
You could also use best of both worlds?

```
-i
--isolated-etcd
```

Just an idea

xetys:
actually, I will need some testing hands when I solve the last problem to make this stuff work.

In the current state, this PR fails at the very last step "join node", because I do a `kubeadm token create` when the API server is still not up after the `apiserver-count` change. I'll need some kind of retryer for that.

There is another very annoying task, which is a little embarrassing for me, but never had a better idea: my progress bars. As of uiprogress I've to set the step count in advance. This means I need to know, how many `AddEvent`s a single progress bar will have, depending on the option. Currently, someone needs to physically count how many steps it actually takes to make an accurate progress meter. With a bunch of new options there is some work to do....god damn UX

pierreozoux:
For the progress bar, I'd keep it simple:
https://github.com/xetys/hetzner-kube/pull/22/files#diff-2acc7cd640c4d0fa151f41f4b35f24e3R155

Instead of counting each steps, just say that each is one step, and it will already be good enough I'd say, and keep the code simple.

You want to count exactly, like with kops?

xetys:
If I count too few, then steps, the event channel for that node closes too early and the tool hangs on the next event emission.

If I count too much, then the bar will move to say 70 - 80 % and immediately jump to 100% when done. It's not that bad like in the first case, but ugly.

This feature is classical UX...makes a lot of work just for the user to be happy. But we know how much UX makes the game in software :)

xetys:
@pierreozoux you actually can start to test if you know how to build this branch from source

xetys:
first bug: isolated mode does not work, yet

xetys:
resolved and read to be merged

SallySwartz:
@xetys tested HA, looks great, it worked like a breeze!
I played with 2 worker nodes, installed rocketchat, killed a node, and well, it didn't work anymore :)

Rook actually needs 3 worker nodes!

xetys:
yeah, with 2 nodes you will have one node with 2 ceph mons on it. If you kill this one ceph cannot maintain the quorum. Maybe adding some kind of warning for the rook addon?

pierreozoux:
I just tried with 3 nodes, killing one worker node, it also failed.
Reading more documentation to try more :)

SallySwartz:
I just tried with 3 nodes, killing one worker node, it also failed.
Reading more documentation to try more :)

PR #25, opened 2018-02-28
fix: send "complete!" event if etcd is deployed isolated

### LINKED ISSUE #24: Process hangs when --isolated-etcd is passed
The tool actually does everything correctly but doesn't state it's done and looks mostly like a bug.

JohnnyQQQQ:
If etcd is deployed on isolated nodes we have to return a "complete!"
event so the deployment process finish succesfully

closes #24


JohnnyQQQQ:
I did a manual check, works now with the flag enabled and without the flag.

xetys:
LGTM

PR #29, opened 2018-03-01
feat: implement the handling of duplicated ssh-keys and ssh-key names

### LINKED ISSUE #23: Adding SSH key fails if the key is already registered with Hetzner
I got a 'uniqueness constraint' violation, and the key wasn't added to hetzner-kube's configuration.

In this case, it should be fine to simply continue setup.

JohnnyQQQQ:
closes #23

### DIFF
--- a/cmd/sshKeyAdd.go
+++ b/cmd/sshKeyAdd.go
@@ -25,6 +25,9 @@ import (
 	"log"
 	"os"
 	"strings"
+	"net/http"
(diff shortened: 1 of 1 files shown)


JohnnyQQQQ:
I did a few manual tests:

- Adding a key that already exists in the local config
- Adding a key that already exists in the hetzner backend
- Adding a new key with an existing name

xetys:
thanks! that's a great UX improvement

Issue #38, opened 2018-03-02
PR #37 seems to introduce a regression that breaks wireguard installation


dmahlow:
Installation with current master:
```
2018/03/02 16:25:21 Creating new cluster staging1 with 1 master(s), 1 worker(s), HA: false
2018/03/02 16:25:21 creating server 'staging1-master-01'...
  10s [====================================================================] 100%
2018/03/02 16:25:38 Created node 'staging1-master-01' with IP <redacted>
2018/03/02 16:25:38 creating server 'staging1-worker-01'...
  15s [====================================================================] 100%
2018/03/02 16:25:59 Created node 'staging1-worker-01' with IP <redacted>
2018/03/02 16:25:59 sleep for 30s...
staging1-master-01   : configure wireguard                 5.0% [--------------]
staging1-worker-01   : configure wireguard                27.3% [==>-----------]
2018/03/02 16:28:54 Failed to execute operation: No such file or directory

2018/03/02 16:28:54 > systemctl enable wg-quick@wg0 && systemctl restart wg-quick@wg0
2018/03/02 16:28:54
2018/03/02 16:28:54
2018/03/02 16:28:54 Run failed:Process exited with status 1
```

systemd log from the worker:
```
Mar 02 16:43:23 staging1-worker-01 wg-quick[9275]: Warning: `/etc/wireguard/wg0.conf' is world accessible
Mar 02 16:43:23 staging1-worker-01 wg-quick[9275]: [#] ip link add wg0 type wireguard
Mar 02 16:43:23 staging1-worker-01 kernel: wireguard: module verification failed: signature and/or required key missing - tainting kernel
Mar 02 16:43:23 staging1-worker-01 kernel: wireguard: WireGuard 0.0.20180218 loaded. See www.wireguard.com for information.
Mar 02 16:43:23 staging1-worker-01 kernel: wireguard: Copyright (C) 2015-2018 Jason A. Donenfeld <Jason@zx2c4.com>. All Rights Reserved.
Mar 02 16:43:23 staging1-worker-01 wg-quick[9275]: [#] wg setconf wg0 /dev/fd/63
Mar 02 16:43:23 staging1-worker-01 wg-quick[9275]: Line unrecognized: `PrivateKey='
Mar 02 16:43:23 staging1-worker-01 wg-quick[9275]: Configuration parsing error
Mar 02 16:43:23 staging1-worker-01 wg-quick[9275]: [#] ip link delete dev wg0
[...]

PR #40, opened 2018-03-02
Fix PR-37 external hcloud manager

mcadam:
Quick fix to just remove the extra flag `--cloud-provider=external` on provisioning external node.
Not fully tested right now as the build is broken, see #38

### DIFF
--- a/cmd/clusterAddExternalWorker.go
+++ b/cmd/clusterAddExternalWorker.go
@@ -15,12 +15,12 @@
 package cmd
 
 import (
-	"github.com/spf13/cobra"
 	"errors"
 	"fmt"
+	"github.com/spf13/cobra"
+	"github.com/xetys/hetzner-kube/pkg"
 	"log"
 	"strings"
-	"github.com/xetys/hetzner-kube/pkg"
 )
 
 // clusterAddWorkerCmd represents the clusterAddWorker command
@@ -60,8 +60,9 @@ An external server must meet the following requirements:
 		}
 
 		externalNode := Node{
-			IPAddress: ipAddress,
+			IPAddress:  ipAddress,
 			SSHKeyName: cluster.Nodes[0].SSHKeyName,
+			IsExternal: true,
 		}
 
 		sshKeyName := cluster.Nodes[0].SSHKeyName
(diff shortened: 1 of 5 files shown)


cmd/nodeUtil.go
```diff
@@ -88,6 +88,8 @@ func (cluster *Cluster) CreateNodes(suffix string, template Node, datacenters []
 }
 
 func (cluster *Cluster) ProvisionNodes(nodes []Node) error {
+	nodes = cluster.Nodes
```
xetys: this is not the right way, as the only time we need to provision all nodes, it's during creation. When adding new workers, this statement will cause unintended side-effects.

cmd/clusterAddExternalWorker.go
```diff
@@ -142,6 +143,8 @@ An external server must meet the following requirements:
 
 		FatalOnError(err)
 
+		cluster.Nodes = append(cluster.Nodes, externalNode)
```
xetys: this statement already comes 4 lines later. I save the new state as soon as the first changing operation succeeded, as ProvisionNodes()

mcadam: Yep just realised how the wait mechanism works, it was not working :) was refactoring to add a proper function and wait for the installation to actually happen, maybe I dont need to anymore 👍

xetys: Yeah you just can remove this one

mcadam:
Ok now we have the behaviour that we want as we don't add the kubelet option to specify external cloud, I also added the option to pass a SSH port on the external node add command. 
Also on the remove command we were not capturing the passphrase so it was not able to delete the node from kubectl directly.

mcadam:
So after recreating a cluster from scratch and trying to add my eternal node, everything works as expected, kubelet is started without `--cloud-provider=external` but still the node gets deleted automatically, so not enough apparently.

xetys:
This is because of the hcloud-controller-manager. 

Until the issue at that project, the only workaround is to delete that deployment. 

You could build a check, if the deployment is present and in that case, remove it before joining the new node

PR #42, opened 2018-03-03
Addon cert manager

lilvinz:
* adds cert-manager as addon which should be able to provide letsencrypt certificates for ingress
* fixes a typo in addon_ingress

### DIFF
--- a/cmd/addon_cert-manager.go
+++ b/cmd/addon_cert-manager.go
@@ -1,0 +1,27 @@
+package cmd
+
+import "log"
+
+type CertmanagerAddon struct {
+	masterNode *Node
+}
+
+func NewCertmanagerAddon(cluster Cluster) ClusterAddon {
+	masterNode, err := cluster.GetMasterNode()
+	FatalOnError(err)
+	return CertmanagerAddon{masterNode: masterNode}
+}
+
+func (addon CertmanagerAddon) Install(args ... string) {
+	node := *addon.masterNode
+	_, err := runCmd(node, "helm install --name cert-manager --namespace kube-system --set ingressShim.extraArgs='{--default-issuer-name=letsencrypt-prod,--default-issuer-kind=ClusterIssuer}' stable/cert-manager")
+	FatalOnError(err)
+	log.Println("cert-manager installed")
+}
+
+func (addon CertmanagerAddon) Uninstall() {
+	node := *addon.masterNode
+	_, err := runCmd(node, "helm delete --purge cert-manager")
+	FatalOnError(err)
+	log.Println("cert-manager uninstalled")
+}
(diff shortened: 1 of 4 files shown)


xetys:
thanks! good job

cmd/addon_cert-manager.go
```diff
@@ -0,0 +1,27 @@
...
+
+func NewCertmanagerAddon(cluster Cluster) ClusterAddon {
+	masterNode, err := cluster.GetMasterNode()
+	FatalOnError(err)
+	return CertmanagerAddon{masterNode: masterNode}
+}
+
+func (addon CertmanagerAddon) Install(args ... string) {
+	node := *addon.masterNode
+	_, err := runCmd(node, "helm install --name cert-manager --namespace kube-system --set ingressShim.extraArgs='{--default-issuer-name=letsencrypt-prod,--default-issuer-kind=ClusterIssuer}' stable/cert-manager")
```
pierreozoux: `extraArgs='{--default-issuer-name=letsencrypt-prod,--default-issuer-kind=ClusterIssuer}'`

If you read carefully the doc, it says that this argument is needed if you want backward compatibility with `kube-lego`. (i didn't read the code to be honest)

I did test without, and it works!
https://www.youtube.com/watch?v=BXHyvPdhEFQ
(Look in the description, there are all the params used.)

pierreozoux:
Also quick question @lilvinz @xetys what do you think about:
 - adding a default ClusterIssuer?
 - passing the email as a param to the command?

Here is a default ClusterIssuer that works:
```
apiVersion: certmanager.k8s.io/v1alpha1
kind: ClusterIssuer
metadata:
  name: letsencrypt
spec:
  acme:
    server: https://acme-v01.api.letsencrypt.org/directory
    email: certificates@example.com
    privateKeySecretRef:
      name: letsencrypt-cluster-key-pair
    http01: {}
```

Then in the ingress, we need to put these annotations:
``` 
annotations: certmanager.k8s.io/cluster-issuer: letsencrypt
```

pierreozoux:
@lilvinz created this PR: https://github.com/xetys/hetzner-kube/pull/45/files

lilvinz:
@pierreozoux I am using exactly that kind of ClusterIssuer and it works for me as well. Shall we add this to the README.md or change something?

cmd/addon_cert-manager.go
lilvinz: I think these are sane defaults. The user is always free to override them. When adding a ClusterIssues like you posted below and name it accordingly "letsencrypt-prod" it works just fine.

pierreozoux:
@lilvinz I believe this should live upstream :)
https://github.com/kubernetes/charts/pull/3975

PR #44, opened 2018-03-03
Addon docker registry

lilvinz:
* adds docker-registry addon and configures it for persistent storage
* requires PVC support in cluster e.g. by Rook

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ You can install some addons to your cluster using the `cluster addon` sub-comman
 * OpenEBS
 * NGinx ingress controller (requires helm)
 * [cert-manager](https://github.com/jetstack/cert-manager) (requires helm)
+* [docker-registry](https://github.com/kubernetes/charts/tree/master/stable/docker-registry) (requires PVC)
 
 ### contributing new addons
 
--- a/cmd/addon_docker-registry.go
+++ b/cmd/addon_docker-registry.go
@@ -1,0 +1,27 @@
+package cmd
+
+import "log"
+
+type DockerregistryAddon struct {
+	masterNode *Node
+}
+
+func NewDockerregistryAddon(cluster Cluster) ClusterAddon {
+	masterNode, err := cluster.GetMasterNode()
+	FatalOnError(err)
+	return DockerregistryAddon{masterNode: masterNode}
+}
+
+func (addon DockerregistryAddon) Install(args ... string) {
+	node := *addon.masterNode
+	_, err := runCmd(node, "helm install --set persistence.enabled=true stable/docker-registry")
+	FatalOnError(err)
+	log.Println("docker-registry installed")
+}
+
+func (addon DockerregistryAddon) Uninstall() {
+	node := *addon.masterNode
+	_, err := runCmd(node, "helm delete --purge docker-registry")
+	FatalOnError(err)
+	log.Println("docker-registry uninstalled")
+}
--- a/cmd/addons.go
+++ b/cmd/addons.go
@@ -7,7 +7,7 @@ type ClusterAddon interface {
 
 func AddonExists(addonName string) bool {
 	switch addonName {
-	case "helm", "rook", "ingress", "openebs", "cert-manager":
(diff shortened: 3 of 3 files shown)


lilvinz:
Hm, it seems a storage class for the PVC has to be set. This needs some more testing before merge.

pierreozoux:
@lilvinz if you use rook, the storage-class should be:

`rook-block`

https://github.com/rook/rook/blob/master/cluster/examples/kubernetes/rook-storageclass.yaml#L18

pierreozoux:
@xetys you see that's why I think it would be easier to be opinionated, then we can create an ecosystem that we know work well together. 

So I do agree with you that addons are not just a wrapper around helm, because, we can customize them to our particular needs. But if we offer too much options, it will increase the complexity, and it will be difficult to maintain.

lilvinz:
@pierreozoux I agree. I was exactly hesitating to pre-define a storage class because there are at least two, which should currently be supported.

So @xetys what is it you would like to see? Shall i set rook by default and note that in the README.md or shall i add an option and force the user to make a choice?

lilvinz:
Another alternative would be to set rook-block or anything else as cluster default. That is then automatically being chosen.

pierreozoux:
Or an other way of solving, make rook and other storage options create the same storage class (like `ssd` ) In this case, it would be easier to consume later on.

xetys:
Anyways, you must pass the storageClassName to the helm chart if you want to consume it. Or your cluster has a default storage class. But here I think it's better not to set up the storage providers as default per default. You could install both, rook and OpenEBS. If they set the default flag, you will have less control over this question. So I think the proper way should be:

1. Install either `addon install rook --set-default` or `addon install openebs --set-default` 
2. Install `addon install docker-registry` should check, if there is a default storage class, and fail if not.

WDYT?

pierreozoux:
I forgot about default storage class :)

Then, IMHO:
 - when the cluster admin installs a storage plugin, it should be the default storage class, if you want, we could have a switch to allow people to not set it as default, but I think most of people want a default storage and just one storage (except for testing, or complicated setup, that this tool shouldn't cover in my opinion)
 - for the addon to check if there is a default storage class, I don't think it should. It should check if there is a storage provider addon, yes.

Actually, I see that this plugin could also consume s3 :) I think also that it would be more cloud native to consume s3 than regular fs. Rook does provide this. But this is maybe out of the scope of this PR.

xetys:
> Actually, I see that this plugin could also consume s3 :) I think also that it would be more cloud native to consume s3 than regular fs. Rook does provide this. But this is maybe out of the scope of this PR.

Yes, definitely. I don't even know how stable the S3 support from rook is.

So as I get, you want if either rook or OpenEBS are installed, they get marked `default`, if the user doesn't pass some kind of `--no-default` flag?

pierreozoux:
@xetys Yes, I think this would make sense.

(I'm working on [that](https://github.com/rook/rook/pull/1545) to have a proper helm chart for the rook cluster, and be able to set it as default easily)

In the mean time, it might be easier to copy the rook-storage to this repo, with the default set, and point the addon to our local file?

WDYT?

xetys:
I would do it the easy way:

```
# kubectl patch storageclass rook-block -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
```

you can either do something like
```
# kubectl get storageclass | grep -v 'NAME' | awk '{print$1}' | xargs kubectl patch storageclass -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
```
 before for removing all other storageclasses default flag in advance. With this, you won't end up having multiple default classes

pierreozoux:
@xetys something like this then? https://github.com/xetys/hetzner-kube/pull/55
:P

xetys:
as #55 is merged, @lilvinz could you please verify it's now working. If everything is alright, please remove the WIP and I merge it

lilvinz:
This works for me although to get basic authentication working another parameter has to be passed to helm. This might be handled in a separate PR though.

PR #45, opened 2018-03-04
Remove unecessary arg from cert-manager

pierreozoux:
This arg is to maintain backward compatibility with kube-lego.

### DIFF
--- a/cmd/addon_cert-manager.go
+++ b/cmd/addon_cert-manager.go
@@ -14,7 +14,7 @@ func NewCertmanagerAddon(cluster Cluster) ClusterAddon {
 
 func (addon CertmanagerAddon) Install(args ... string) {
 	node := *addon.masterNode
-	_, err := runCmd(node, "helm install --name cert-manager --namespace kube-system stable/cert-manager")
+	_, err := runCmd(node, "helm install --name cert-manager --namespace ingress stable/cert-manager")
 	FatalOnError(err)
 	log.Println("cert-manager installed")
 }
--- a/cmd/addon_cert-manager.go
+++ b/cmd/addon_cert-manager.go
@@ -14,7 +14,7 @@ func NewCertmanagerAddon(cluster Cluster) ClusterAddon {
 
 func (addon CertmanagerAddon) Install(args ... string) {
 	node := *addon.masterNode
-	_, err := runCmd(node, "helm install --name cert-manager --namespace kube-system --set ingressShim.extraArgs='{--default-issuer-name=letsencrypt-prod,--default-issuer-kind=ClusterIssuer}' stable/cert-manager")
+	_, err := runCmd(node, "helm install --name cert-manager --namespace kube-system stable/cert-manager")
 	FatalOnError(err)
 	log.Println("cert-manager installed")
 }


xetys:
as of #46, could you please evaluate, if there are any concern by putting this helm chart into a distinct "cert-manager" namespace, rather than "kube-system", and if that works, integrate that into your PR (by changing the --namespace option in the command you already changed)?

pierreozoux:
@xetys should we put nginx in the ingress ns too?

xetys:
Putting nginx ingress to the proper namespace can be done in a distinct PR. 

If you remove that flag, we should then do a configuration similar to what you are doing in that video. Just by removing the flag, the user will end up with an incomplete setup. So both, documentation and a preconfiguration is needed here. I'm even thinking about enabling kube-lego backward compatibility, as the most other helm charts are currently built with support for that.

pierreozoux:
I'm not sure I fully understand your comment:






> If you remove that flag, we should then do a configuration similar to what you are doing in that video.

You mean the annotation in the `ingress`?



> Just by removing the flag, the user will end up with an incomplete setup.

This is not true, the flag is to be backward compatible with lego, but if we take the example of [rocket.chat](https://github.com/kubernetes/charts/blob/master/stable/rocketchat/values.yaml#L91) there is still configuration to do.
So it is the same as now.

> So both, documentation and a preconfiguration is needed here.

I agree with documentation, but maybe, this should be part of:
https://github.com/kubernetes/charts/blob/master/stable/cert-manager/templates/NOTES.txt
?

> I'm even thinking about enabling kube-lego backward compatibility, as the most other helm charts are currently built with support for that.

`kube-lego` is being deprecated, so we shouldn't rollback, but forward fix. Anyway, currently charts are usually not configured for tls, it has to be a user action to activate it. And so part of this activation, the user should specify cert-manager.

xetys:
> If you remove that flag, we should then do a configuration similar to what you are doing in that video.

You first install the cert-manager, then add a cluster issue over new resource `ClusterIssuer`. These steps are not done during this PR. If you remove the ingressShim, you should then copy the cluster-issuer manifest to the cluster, including the provided email.

> kube-lego is being deprecated

I'm not suggesting installing kube-lego. But setup cert-manager with compatibility. A lot of charts still provide `kubernetes.io/tls-acme: "true"` instead of the new annotation. With that flag enabled, people can use the cert-manager with the same experience. I would keep this until kube-lego really runs out of EOL and cert-manager gets some maturity, with the majority of other charts doing the new way.

So I think, return the ingress-shim flag, install a cluster-issuer like you did for rocket-chat, and provide some docs about it. If this is too much for you, tell me, I can merge any state of this PR and finish that.

pierreozoux:
I don't remove the ingressShim, it is installed by [default](https://github.com/kubernetes/charts/blob/master/stable/cert-manager/values.yaml#L38).

I believe the cluster issuer should live in the [cert-manager chart](https://github.com/kubernetes/charts/pull/3975).

I understand what you say about `kube-lego`. I don't have strong opinion on that, I let you decide what is best.

If you want we can wait the mentioned PR to be merged before merging this change?
(I prefer fix things upstream, I believe this is needed for everybody, not just us. It takes more time, but then it is fixed for real)

xetys:
Ok, I see. So just for my understanding, if I install it in this state with the removed flag, what will happen if I provide an ingress resource with the `kubernetes.io/tls-acme: "true"` annotation, instead of the new cluster-issuer one.

What I understand is, we just could use the new way out of the box, without real backward compatibility.

I need this to be merged as of #46 and to start working on #47

pierreozoux:
> Ok, I see. So just for my understanding, if I install it in this state with the removed flag, what will happen if I provide an ingress resource with the kubernetes.io/tls-acme: "true" annotation, instead of the new cluster-issuer one.

Yes, this will not provision a cert.

Then, let' close this PR, and just fix the NS?

xetys:
I was thinking of how to solve this PR and end up with this: as the cert manager is still broken because of the namespace issue, I will accept this PR. This also means we have the cert manager running in the new way, which differs from the kube-lego way. I whish some sort of legacy mode, where the tool is working like kube-lego, as of the points stated above. I think this mode is a topic for a new PR. So LGTM

PR #55, opened 2018-03-05
Make rook the default storage when installed

pierreozoux:
It feels a bit hacky, but should work


lakano:
Hello @pierreozoux !

Could you tell me which storageClass value we need to set please? ( "", "standard", "default", "." ) 

I've just tried to deploy my own Helm Chart on the HA cluster, but all nodes are in Pending state, and the « kubectl get events » isn't very easy to follow ( there is events in top, bottom, and I don't undestood how to get a list of events ordered by date... ) but I can see with my tests it's related to the name of the StorageClass I've used.
I can see some « PersistentVolumeClaim is not bound », « storageClass.storage.k8s.io "standard" not found » (so it's not "default" ok.. ) 

Thanks!

Issue #47, opened 2018-03-04
Code structure redesign


xetys:
Currently, this project has a code coverage of ~11% and the working parts are coupled tight to each other. This also makes testing hard. As this is my first golang project, it's hopefully ok that I started in this state. But I think now is the time to make it better. For the next minor version, I see a huge improvement on the code quality by:

- [x] move the core part of provisioning the cluster (in `nodeUtil.go`, `config.go`) into a `cluster-deploy` package under `pkg`
- [x] create interfaces for providing `runCmd`, `writeFile` and a source for `Node`, like the hetzner server create stuff. Doing a direct implementation with an SSH client for node communication and the `HetznerConfig` for providing nodes. With this change, the cluster-deploy logic can be provided with mock implementations to test this logic wider
- [x] use the actual commands under `cmd` just for calling these tools. Here we need tests, which prove correct working validations

This enables to write better tests, and ensure the quality of hetzner-kube in a long run. 

I will wait, until #45, #44, and #40 are resolved and do a 0.2.2 release. After this I will start my work on this redesign and for this time stop merging incoming PRs (or these will have to merge the redesign in advance).

JohnnyQQQQ:
I would also suggest to setup Codeclimate for the project so we can track the technial dept sitting around. If you want I can make a PR for the setup.

xetys:
The same for godep...we should fix the versions of the libraries we use to avoid breaking builds because of incompatible vendor stuff.

I would really appreciate a PR for Codeclimate :+1:

JohnnyQQQQ:
I opened one PR regarding both points. We should also take into consideration to checkin the vendor directory.

https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory

xetys:
update, the lion part of this issue was solved weeks ago. What's left is writing tests for the validation. This can be done in further versions/

So I close this ticket as of #75 and #48 and a lot more stuff is already done

PR #75, opened 2018-03-14
[WIP] Code Redesign

xetys:
As stated in #47 here comes the next big thing. It contains a huge package redesign, decoupling the kubernetes installation from cobra and hetzner. The PR also adds the vendor lib to the source, as I want to ensure build stability and reproducibility


JohnnyQQQQ:
@xetys If I can help you in any way to get this PR landed just drop me a line

xetys:
Feel free to help. I already did a heavy lifting part and I tested the basic cluster create runs. But there are a lot of issues with the different sub-commands. As my time is low again, I would be happy to any help.

Even adding more tests would be a great help. For example, we could mock the NodeCommunicator interface with a pseudo-ssh client, to do integration tests for the cluster manager.

xetys:
so, I think this is a good state for now. I tested the common scenarios. The test coverage slightly increased, but codeclimate is telling a lot of issues. 

@JohnnyQQQQ could you please test the current state in this branch before I merge?

JohnnyQQQQ:
@xetys when creating a cluster I can't input the ssh-key passphrase, it seems to be buggy.

JohnnyQQQQ:
buggy_input

It seems like the change of asking first for the passphrase when doing something that is passphrase related is missing.

xetys:
OK, this also working now, so I think it's time to merge this

PR #76, opened 2018-03-14
Update flannel url to use master branch to enable hostPort support

Doccrazy:
This updates the flannel installation URL to https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml, as documented on https://github.com/coreos/flannel. 

This includes the changes from https://github.com/coreos/flannel/pull/888, which enables the CNI portmap plugin to allow hostPort mappings, as required e.g. for the Traefik helm chart. 
Flannel is hard to upgrade in an existing cluster, as config files need to be manually removed and a reboot is required.

### DIFF
--- a/cmd/nodeUtil.go
+++ b/cmd/nodeUtil.go
@@ -272,7 +272,7 @@ func (cluster *Cluster) InstallMasters() error {
 	commands := []SSHCommand{
 		{"kubeadm init", "kubeadm init --config /root/master-config.yaml"},
 		{"configure kubectl", "rm -rf $HOME/.kube && mkdir -p $HOME/.kube && cp -i /etc/kubernetes/admin.conf $HOME/.kube/config && chown $(id -u):$(id -g) $HOME/.kube/config"},
-		{"install flannel", "kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/v0.9.1/Documentation/kube-flannel.yml"},
+		{"install flannel", "kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml"},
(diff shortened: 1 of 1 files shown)


pierreozoux:
Undertand, but still, I think this would be better to point to one version, like `v0.10.0`, and update this version regularly.

What do you think?

Doccrazy:
My understanding is that the flannel config file (from Documentation folder, not directly from source) is explicitly updated by the maintainers on new releases to refer to the latest stable version. 
See e.g. https://github.com/coreos/flannel/blob/c0648a8ff42002d7f3ab2292f79e07584e2bd8a0/Documentation/kube-flannel.yml#L107 and the commit history of that file.

pierreozoux:
Yes, this is the problem, we don't want to ship a new version without knowing it explicitly, this is a good practice.

Doccrazy:
I don't really care which version is used. 0.10.0 includes the portmap fix, so this is fine by me.

xetys:
Hey! Good hint. Could you please open this PR for the redesign branch?

Doccrazy:
Sure, I updated the PR (first time using this feature, yay!)

xetys:
hey, thanks for your PR.

The redesign is current WIP and one of the parts is totally making `nodeUtil.go` obsolete. The new stuff happens in `pkg/cluster/manager.go`. It's logically the same. Could you please fix this last point? Sorry for confusing

mavimo:
@xetys I think we can close this issue

mavimo:
@xetys @Doccrazy @pierreozoux there is something missing from the #124 or we can close this PR?

mavimo:
@xetys @Doccrazy @pierreozoux 🆙

mavimo:
I think we can close it, open it again if I'm wrong, thx!

PR #89, opened 2018-03-30
feat: add the possibility to set the install file url

JohnnyQQQQ:
This can be useful if the user want to hook into the installtion. I used it to make kube-prometheus work, as I needed to append some arguments to the kubelet.

### DIFF
--- a/cmd/cluster_add_worker.go
+++ b/cmd/cluster_add_worker.go
@@ -19,12 +19,12 @@ import (
 	"fmt"
 	"github.com/spf13/cobra"
 	"github.com/xetys/hetzner-kube/pkg"
+	"github.com/xetys/hetzner-kube/pkg/clustermanager"
 	"github.com/xetys/hetzner-kube/pkg/hetzner"
 	"log"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


xetys:
Hey,

I know what you are trying to do and I'm not sure if I like this approach. My thoughts:

- you need this as you can set some flags during setup for kube-prometheus to work. As of https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/docs/kube-prometheus-on-kubeadm.md you can do this after the setup, too. So there is no need to change the routine of it. 
- I actually doesn't really like my current approach with wgetting the file from github at all. That means in particular there is no guaranty of earlier versions still work well with the current script. It's just a lot of stuff to do and I was lazy to do this in a similar fashion to NodeCommand on master setup. 

Actually, for prometheus, I would prefer to do an addon which performs the steps in that doc. In a next step I would like to have a change in the install method by transfering all these steps into a routine inside the tool. And while doing this, we could add some logic like the HetznerProvider is injection the hcloud-controller commands into the default routine.

JohnnyQQQQ:
I had the same thoughts about it, mostly regarding your second point. I just thought it would be handy to have a workaround for this till the second point lands in a release.

PR #90, opened 2018-03-30
cluster create enhanced documentation

kabudu:
I felt that more detailed documentation was required for creating a cluster, especially around the available option flags. I know that a user of the tool could simply look at the code to work this out, but personally, I prefer to have good documentation available. (The challenge will be maintaining it! ;))

I'm not sure whether you have any conventions for how to structure documentation for this tool, but at least this can be a discussion point if it needs to be updated to conform to your standards.


xetys:
thx a very lot!

Issue #104, opened 2018-04-23
move install-docker-kubeadm.sh into code


xetys:
to make sure the tool always use the commands for installation corresponding to its version. If not, the following might happen:

as currently in version 0.3, the tool fetches the install-docker-kubeadm.sh from the repo. If new stuff is invented in 0.4, all users with the older version might not be able to use the tool. 

There are far more reasons for this, e.g. if some new stuff is developed in branches.

xetys:
related to #89

Issue #33, opened 2018-03-01
Hybrid clusters


Baughn:
Still in the 'playing around' mode. One thing I'd like to do is build a cluster backed mostly by my own (already paid-for, fairly hefty) hardware, but with the possibility of adding worker nodes running on Hetzner on-demand.

The best way to do this, I think, would be a Kubernetes daemon that watches for overall load, creating new worker nodes when it reaches some threshold. It would make sense to use hetzner-kube for the heavy lifting.

That much should already be doable; what I'd like to ask for is the ability to link hetzner-kube to a pre-existing (non-Hetzner) cluster, and add workers to that.

There already is support for adding workers to existing clusters, so while this might be somewhat out of scope for the project, *hopefully* it wouldn't be too difficult to add. If you'd rather not, that's fine; I can probably hack something up. From the looks of config.json it might not even require any new code, but I'd like to confirm that.

(One likely stumbling block is Wireguard, which I've yet to try setting up myself. Should be fun.)

SallySwartz:
I think it is out of scope of the project, what you are ooking for is an autoscaler if I understood well your question.

Look at this for instance:
https://github.com/kubernetes/charts/tree/master/stable/aws-cluster-autoscaler

We could imagine the same for hetzner.

xetys:
Two thoughts on that:

1. Indeed, this project contains already some general kubernetes context which is interesting for kubernetes cluster in general. Technically this tool list tightly coupled with the hetzner API. But even that is just for the beginning part. After that the only thing needed is there IP of the server for SSH. 
I'm already thinking of some kind of refactoring, to make the actual provision part independent of the API, and works with anything what can provide server info. A yaml file, for example. This is a lot of work but enables this tool to be used in a general context.

2. You are asking for an option to provision an existing server the same way as hetzner generated. A straightforward way to do that is to add a "add-external-worker" command with an additional flag "IP" and the assumption, it has Ubuntu 16.04 with SSH available with the same key as used for the cluster. This means to build a check, which validates the provided IP to satisfy these requirement and afterwards perform the same steps like before

That is not too much work I think

mcadam:
I would also be interested to connect current machines to a new cluster created on Hetzner. I think the road to provide an ip and a ssh key to make the connection and then apply the same scripts as the Hetzner worker ones would be nice and generic enough.

xetys:
```
# kubectl get node 
NAME                   STATUS    ROLES     AGE       VERSION
[...]

PR #105, opened 2018-04-23
move hcloud-controller-manager to addon, improve installation

xetys:
This is a double PR which

## 1. moves hcloud-controller-manager to the addons section

So it is not installed by default anymore! This is in favor of #33 and #40. If you are going to use hcloud only nodes, run `hetzner-cloud cluster addon install hcloud-controller-manager -n ...` to install the manager. Note, that adding external workers won't work then.

## 2. moves package install routine from the script into code

This gives the progress bars a better experience and of course, mainly solves problems mentioned in #104 and #89


xetys:
the failing stuff looks like a temporary travis isse, merging then

PR #125, opened 2018-05-12
Fix golint issues

mavimo:
- [x] Fix all `golint` issues from codebase
 - [x] Introduce `golint` as CI step


mavimo:
@xetys I fixed 81 issues reported from codeclimate (related to `golint`) but there are 4 new issues that I'm not sure on how to solve; you can see it on [codeclimate](https://codeclimate.com/github/xetys/hetzner-kube/pull/125) (clear filters and flag only "new"). 
Reported errors indicate "duplicate code" but I'm not sure we should remove the reported duplication without impact on code simplicity (and I'm not sure we should fix the reported issues), can you take a look? Thanks!

mavimo:
@xetys ok, I should stop to do the suggested refactor because more I refactor more issue are detected by codeclimante 😄

xetys:
Haha, I know, I faced that in all my previous PRs. Even if I have the power
to just press the "accept" button in code climate, I just fix it to improve
the maintainability with every update. So don't be afraid, just keep on
track, you're doing a great job!

That's important as of version 0.3 the kubernetes logic is generalized, so
you can use this tool as a library in different projects. This is why the
code quality matters so much for me

Marco Vito Moscaritolo <notifications@github.com> schrieb am Sa., 12. Mai
2018, 11:46:

mavimo:
@xetys np ;)

PR #144, opened 2018-06-08
fix: add external worker

JohnnyQQQQ:
Fixes also the progressbar for internal and external worker


xetys:
Coding from smartphone is such a pita

JohnnyQQQQ:
@xetys fixed last bug

PR #154, opened 2018-06-21
[WIP] feat: use ubuntu 18.04

JohnnyQQQQ:
I think this is a good step for version 1.0 as it fixes alot of problems comming with the old kernel version etc.

I will add some integration tests that we need.

### DIFF
--- a/pkg/clustermanager/provision_node.go
+++ b/pkg/clustermanager/provision_node.go
@@ -10,7 +10,7 @@ import (
 const maxErrors = 3
 
 // K8sVersion is the version that will be used to install kubernetes
-var K8sVersion = flag.String("k8s-version", "1.9.6-00",
+var K8sVersion = flag.String("k8s-version", "1.10.5-00",
 	"The version of the k8s debian packages that will be used during provisioning")
 
 // NodeProvisioner provisions all basic packages to install docker, kubernetes and wireguard
@@ -51,7 +51,8 @@ func (provisioner *NodeProvisioner) Provision(node Node, communicator NodeCommun
 	}
 
 	eventService.AddEvent(node.Name, "packages installed")
-	return nil
+
+	return provisioner.disableSwap()
 }
 
 func (provisioner *NodeProvisioner) packagesAreInstalled(node Node, communicator NodeCommunicator) bool {
@@ -82,6 +83,18 @@ func (provisioner *NodeProvisioner) prepareAndInstall() error {
 	}
 
 	return nil
+}
+
+func (provisioner *NodeProvisioner) disableSwap() error {
+	provisioner.eventService.AddEvent(provisioner.node.Name, "disabling swap")
+
+	_, err := provisioner.communicator.RunCmd(provisioner.node, "swapoff -a")
+	if err != nil {
+		return err
+	}
+
+	_, err = provisioner.communicator.RunCmd(provisioner.node, "sed -i '/ swap / s/^/#/' /etc/fstab")
+	return err
 }
 
 func (provisioner *NodeProvisioner) installTransportTools() error {
@@ -139,7 +152,7 @@ func (provisioner *NodeProvisioner) prepareDocker() error {
 	// docker-ce
 	aptPreferencesDocker := `
 Package: docker-ce
-Pin: version 17.03.*
+Pin: version 18.03.*
 Pin-Priority: 1000
 	`
 	err := provisioner.communicator.WriteFile(provisioner.node, "/etc/apt/preferences.d/docker-ce", aptPreferencesDocker, false)
--- a/pkg/hetzner/hetzner_provider.go
+++ b/pkg/hetzner/hetzner_provider.go
@@ -63,7 +63,7 @@ func (provider *Provider) CreateNodes(suffix string, template clustermanager.Nod
 			Name: template.Type,
 		},
 		Image: &hcloud.Image{
-			Name: "ubuntu-16.04",
+			Name: "ubuntu-18.04",
 		},
 	}
 


pkg/clustermanager/provision_node.go
```diff
@@ -139,7 +152,7 @@ func (provisioner *NodeProvisioner) prepareDocker() error {
 	// docker-ce
 	aptPreferencesDocker := `
 Package: docker-ce
-Pin: version 17.03.*
+Pin: version 18.03.*
```
eliasp: This shouldn't be changed. This was used to pin the Docker CE release to prevent automatic upgrades to newer versions which aren't compatible with K8S.

See also: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.10.md#external-dependencies

JohnnyQQQQ: I'm aware of that, for now 18.03 is the only version in the stable repository from docker ce afaik. Perhaps we should use the docker.io package, I think it's on 17.03

eliasp: - The Ubuntu/bionic provided [docker.io package](https://packages.ubuntu.com/bionic/docker.io) is at 17.12.1, no older versions available :red_circle: 
- The Docker CE package is at [18.03](https://download.docker.com/linux/ubuntu/dists/bionic/stable/binary-amd64/Packages) :red_circle:

None of those versions is compatible.

Another option would be to switch to [cri-o](http://cri-o.io/) and thereby finally getting rid of Docker…

xetys:
I have no experience with cri-o and don't know if it's that stable as
docker is. I don't want to make experiments as a more and more users use
this tool with defaults. If it's bleeding edge it's probably not the best
idea.

About docker install, they're these install scripts from rancher which
always did a good job for me do far. Maybe use them?

Elias Probst <notifications@github.com> schrieb am Fr., 22. Juni 2018,
12:14:

JohnnyQQQQ:
@xetys I found this Dockerfile https://github.com/rancher/rancher/blob/040665c4b523243e54a9c5c3f22528f1998ce5fa/package/Dockerfile.agent

I will use try this approach

pkg/clustermanager/provision_node.go
qubusp: @JohnnyQQQQ docker-ce 17.03 should be available in the repo cause this bellow works. 
```
export docker_version=17.03.2
apt-get install -y -q docker-ce=$(apt-cache madison docker-ce | grep ${docker_version}
OR
install -y -q --setopt=obsoletes=0 docker-ce-${docker_version}.ce
```

JohnnyQQQQ: @qubusp I still only get versions > 17

JohnnyQQQQ: I have subscribed to https://github.com/kubernetes/kubernetes/issues/65089 and will wait for any updates

qubusp: K8s 1.10 supports Docker 17.3.x if I'm not wrong: https://kubernetes.io/docs/imported/release/notes/
`https://kubernetes.io/docs/imported/release/notes/`

xetys: You can install docker 17.03 on ubuntu 18.04 using snap:

``` bash
snap install docker --channel=17.03/stable
```

xetys: no, it's not a good idea to use snap. I can't run privileged containers and it's not running in systemd.

About cri-o, there is also no support for 18.04, yet. So this won't help us in this problem anyways. 

I fear we need to get up a stable runtime first, in a supported way. That looks somehow really complicated at the moment :disappointed:

galexrt:
@xetys CRI-O is pretty stable for me. I have been running it for more than two months now in my cluster.

@JohnnyQQQQ Why not use Fedora? Fedora is even more up-to-date from kernel and/or other components (e.g. the Fedora repos contain a Docker package AFAIK). :slightly_smiling_face:

pkg/clustermanager/provision_node.go
```diff
@@ -10,7 +10,7 @@ import (
 const maxErrors = 3
 
 // K8sVersion is the version that will be used to install kubernetes
-var K8sVersion = flag.String("k8s-version", "1.9.6-00",
+var K8sVersion = flag.String("k8s-version", "1.10.5-00",
```
gadelkareem: Not sure if this is still in progress but version 1.12 is out

xetys:
hey,

I tested some stuff right now. You can install k8s 1.11+ on ubuntu 18.04 if using the containerd runtime of docker directly, which is available since version 18.06. I tried the following:

current version of kubeadm etc. following master config:

```
apiVersion: kubeadm.k8s.io/v1alpha3
kind: InitConfiguration
nodeRegistration:
  criSocket: /var/run/docker/containerd/docker-containerd.sock
  name: docker-test
  taints:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
---
apiVersion: kubeadm.k8s.io/v1alpha3
kind: ClusterConfiguration
networking:
  serviceSubnet: "10.96.0.0/12"
  podSubnet: "10.244.0.0/16"
  dnsDomain: "cluster.local"
```

and install canal as network adapter.

Finally this is my output of node description:

```
Name:               docker-test
Roles:              master
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/os=linux
                    kubernetes.io/hostname=docker-test
                    node-role.kubernetes.io/master=
 #...
 Kernel Version:             4.15.0-34-generic
 OS Image:                   Ubuntu 18.04.1 LTS
 Operating System:           linux
 Architecture:               amd64
 Container Runtime Version:  containerd://1.1.1
 Kubelet Version:            v1.12.1
 Kube-Proxy Version:         v1.12.1
PodCIDR:                     10.244.0.0/24
```

as you see, k8s takes the containerd runtime provided. That's a stable way tu run k8s on ubuntu 18.04, I think

pkg/clustermanager/provision_node.go
xetys: referring to my last comment, we need to pin 18.06 

and use the config I mentioned lower

JohnnyQQQQ:
closing in favour of #208

PR #157, opened 2018-06-25
feat: add master-ip command

JohnnyQQQQ:
To have an easy way to automate things, it's handy to have a command that will return the master-ip of the cluster so one can fetch the kubeconfig or something else.

### DIFF
--- a/cmd/cluster_master_ip.go
+++ b/cmd/cluster_master_ip.go
@@ -1,0 +1,58 @@
+// Copyright © 2018 NAME HERE <EMAIL ADDRESS>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


JohnnyQQQQ:
usage example 

```bash
MASTER_IP=$(hetzner-kube cluster --name production-new master-ip)
scp -i ~/some_key root@$MASTER_IP:~/.kube/config cluster_config
```

xetys:
looks good. But could you please make it like the "new" standard, which is more convenient with tools like kubectl, kubeadm etc:

```
hetzner-kube cluster production-new master-ip
```
using args, instead of

```
hetzner-kube cluster --name production-new master-ip
```

JohnnyQQQQ:
Sure, I'll update it later

JohnnyQQQQ:
@xetys is there an example were we already do this? Only found some where the name is a leaf argument (at the end of the command not before)

xetys:
sry, I mean of course

```
hetzner-kube cluster master-ip production-new 
```

like in `hetzner-kube kubeconfig my-cluster`

JohnnyQQQQ:
@xetys I think the refactoring that codeclimate suggest is worthless for now, you will still have to do the same checks if you wrap it an a function, just with an error instead of an empty string.

PR #159, opened 2018-07-09
Add info about Docker image to README

bittner:
I discovered that there is a Docker image for `hetzner-kube`. This is probably worth noting in the README.

P.S.: Can you disclose where the sources for this image are hosted? They don't seem to be on GitHub.


bittner:
That's sounds like a problem we can resolve. Why don't you simply add the Dockerfile you used back then to this repository, and we set up automatic builds on Docker Hub?

Issue #122, opened 2018-05-09
e2e suite


xetys:
issues like #99 and #119 and some similar shows the need of a sustained way to test this tool works properly. For this purpose I suggest creating an e2e suite, which takes an API token for a hetzner account and tests these scenarios:

1. `cluster create -n level-one -k XXX -w 1` (creates 2 nodes) 
2. add worker (adds one more node)
3. `cluster create -n level-three -k XXX --ha-enabled -w 1` (creates 4 nodes)
4. `cluster create -n level-four -k XXX --ha-enabled --isolated-etcd` (creates 5 nodes)

This would lead to 12 nodes created per  CI run. This would cost me 6 cent, I think its worth for hetzner-kube to become a better tool...

I will try to realize this in the next days

schliflo:
Maybe it's even worth asking Hetzner to sponsor an account for CI Testing purposes. I mean in the end this tool probably generates a fair amount of income for them already.
Pinging random people from [hetznercloud](https://github.com/orgs/hetznercloud/people) @thcyron @thetechnick

JohnnyQQQQ:
Are there any news on this? I have some basic e2e tests in bash ready to go..

xetys:
I was busy the past time so didn't really started here yet. Feel free to
share ideas

Jean-Philippe Quéméner <notifications@github.com> schrieb am So., 3. Juni
2018, 18:02:

JohnnyQQQQ:
I would write a shell script that would run all three tests above and clean up the account after every test. On failure it would return an error exitcode.

mavimo:
@xetys @schliflo @JohnnyQQQQ @voron I start to create the E2E test suite (actually a PoC) that create context, add ssh-key (that are free operation on hetzner cloud) and I'm planning to introduce cluster creation, scaling and destroy.
You can find the WIP on https://github.com/mavimo/hetzner-kube/commits/introduce-e2e-tests (actually work as expected, but there are no assertion on task).

I'm investigating https://github.com/sstephenson/bats as test tool for bash script, any suggestion on different tools that can be used?

PS: I renamed project to mavimo/hetzner-cloud or travis will not compile the code, I'll revert before opening the PR (this operation involve some admin operation on Travis-CI, @xetys can I write the instruction for config via mail or ...?)

mavimo:
@xetys do you think we can close it and open more specific issue for the "next steps" ?

xetys:
[...]

PR #174, opened 2018-08-16
Introduce e2e tests

mavimo:
Refs #122

First tentative to introduce end-to-end test, currently only check that the command are invoked correctly and do not return issue.
This branch include also the fix reported in https://github.com/xetys/hetzner-kube/pull/173 used to set token during context creation.

@xetys @JohnnyQQQQ this PR is not mergeable but is a starting point to discuss how to setup an appropriate E2E test suite

NB: PR fail since need config on travis-CI side, you can see the build working on a configured env in: https://travis-ci.org/mavimo/hetzner-kube/branches branch `introduce-e2e-tests`


xetys:
That's so really amazing, as I still didn't find any time do this very important mission!

I will first take a closer look at your previous PRs before digging into this one, but that is awesome!!!

mavimo:
@xetys the build is failing since the `HETZNER_API_KEY` is missing from travis config, once it is set you should se a build correctly running 😄 

Eg: https://travis-ci.org/mavimo/hetzner-kube/builds/438378321

xetys:
we try to merge it to solve the issue with the key by merging that

PR #175, opened 2018-08-17
Fix external worker private IP choice

voron:
Just hit a bug while adding external worker `test1-eworker-01` - it uses already taken private IP, and thus alive node `test1-worker-05` goes to `NotReady`
```
cat .hetzner-kube/config.json|jq '.clusters[0].nodes[].name,.clusters[0].nodes[].private_ip_address'
"test1-master-01"
"test1-worker-01"
"test1-worker-02"
"test1-worker-05"
"test1-worker-07"
"test1-eworker-01"
"10.0.1.1"
"10.0.1.21"
"10.0.1.22"
"10.0.1.25"
"10.0.1.27"
"10.0.1.25"
```
And yes, I removed some workers by hand. 

Here is a quick fix to find first free private IP

### DIFF
--- a/cmd/cluster_add_external_worker.go
+++ b/cmd/cluster_add_external_worker.go
@@ -123,14 +123,19 @@ An external server must meet the following requirements:
 		FatalOnError(err)
 		externalNode.Name = hostname
 
+		cidrPrefix := clustermanager.PrivateIPPrefix(cluster.NodeCIDR)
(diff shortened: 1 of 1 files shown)


JohnnyQQQQ:
LGTM, can you apply gofmt to the changed file pleased. Otherwise codeclimate will not pass.

Issue #139, opened 2018-06-01
Routing of Flannel traffic over wireguard


jowenn:
It is nice that there is wireguard, but it is not used for inter pod communication

Flannel by default uses the  device with the default route (if run on the host), I'm not sure how it is handled when run as a pod

At least I could verify that my communication between two pods on two nodes is not encrypted, because it does not use the wg0 interface.

What I did:
I created a cluster
hetzner-kube cluster create ..... --ha-enabled -w 3

I launched 3 nginx containers  (each container is scheduled on one node)
kubectl run nginx1 --image=nginx --replicas=3 --port=80

I started on each node (directly)
tcpdump -nnvvSSs 1514 -i eth0 | grep GET

I ran a single interactive container with curl:
kubectl run curl1 --image=radial/busyboxplus:curl -i -t --rm
In this container I startet
curl http://10.244...../MYTEST1 >/dev/null

For every connection between the busy box and a container on another node I gut a nice
GET /MYTEST1 HTTP/1.1 output from tcpdump
 
Therefore the traffic is no going flannel->wg0->encryption->eth0->and on but is going directly flannel->eth0, which means the traffic is not encrypted.

I also got some other GET /  for instance when kubectl connected a terminal to a pod, so it appears that also not all control plane traffic is encrypted, except etcd, which seems to really use wg0

pierreozoux:
Yes, noticed that too.

I think what is happening is that wireguard is created, then kubeadm is configured to use wireguard IPs, and then when flannel starts, it discovers interface on its own, and choose eth.

I had to deploy my cluster with kubeadm (additional worker failing), and I used that config for flannel on top of wireguard:
https://git.indie.host/indiehost/standard/blob/master/kube-flannel.yml#L74-85

Hope it helps!

jowenn:
I've tried removing everything flannel related and used your file with kubectl apply and with kubectl create in both cases the result is that the pods are in crash loop. From the logs I can see that the post startup command errors out with a "Address already in use" message

pierreozoux:
Yeah, maybe you have to clean some folders too, and iterfaces.

I've used this command when I used kubeadm, to reset node:
```
systemctl stop kubelet
systemctl stop docker
rm -rf /var/lib/cni/
rm -rf /var/lib/kubelet/*
rm -rf /etc/cni/
ifconfig cni0 down
ifconfig flannel.1 down
ifconfig docker0 down
```
[...]

PR #178, opened 2018-08-25
Route flannel via existing wireguard interface

voron:
Untested
Closes [#139](https://github.com/xetys/hetzner-kube/issues/139)

### DIFF
--- a/pkg/clustermanager/cluster.go
+++ b/pkg/clustermanager/cluster.go
@@ -153,6 +153,7 @@ func (manager *Manager) InstallMasters() error {
 		{"configure kubectl", "rm -rf $HOME/.kube && mkdir -p $HOME/.kube && cp -i /etc/kubernetes/admin.conf $HOME/.kube/config && chown $(id -u):$(id -g) $HOME/.kube/config"},
 		{"install flannel", "kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/v0.10.0/Documentation/kube-flannel.yml"},
 		{"configure flannel", "kubectl -n kube-system patch ds kube-flannel-ds --type json -p '[{\"op\":\"add\",\"path\":\"/spec/template/spec/tolerations/-\",\"value\":{\"key\":\"node.cloudprovider.kubernetes.io/uninitialized\",\"value\":\"true\",\"effect\":\"NoSchedule\"}}]'"},
+		{"configure flannel to work via wireguard interface", "kubectl -n kube-system patch ds kube-flannel-ds -p '{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"args\":[\"--ip-masq\",\"--kube-subnet-mgr\",\"--iface\",\"wg0\"],\"name\":\"kube-flannel\"}]}}}}'"},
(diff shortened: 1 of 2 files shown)


voron:
@xetys @mavimo @JohnnyQQQQ pls advise how to pass codeclimate 4 return statements  limit

xetys:
The latter option is the proper way to do that I think. The motivation
behind that comes to mind when someone reads the code for the first time.
Reading two functions with less return statements will be easier to
understand than one function with more. It feels like codeclimate is a
little bit Nazi on that, but the idea behind that still makes sense.

So I suggest to split that into two functions and then we need to test it,
as you stated it is not tested yet. After that I think we can merge it.

Marco Vito Moscaritolo <notifications@github.com> schrieb am So., 26. Aug.
2018, 15:24:

mavimo:
@xetys IMHO this rule is too strict in Golang context where the early return approach is a standard, especially for error returning.

Aside the above opinion, on the previous comment my last proposal was:
```go
func (provisioner *NodeProvisioner) preparePackages() error {
	// ...
	err = provisioner.prepareNetwork()
	if err != nil {
		return err
	}

	return nil
}

func (provisioner *NodeProvisioner) prepareNetwork() error {
	err = provisioner.prepareFlannel()
	if err != nil {
		return err
	}

	err = provisioner.prepareWireguard()
	if err != nil {
		return err
	}
	return nil
}
```
but after consideration I think we should invoke `prepareNetwork` in the main function and not into `preparePackage` function, since is not a package but a service configuration, @voron @xetys WDYT?

voron:
@mavimo @xetys I tried to implement some function separation, but I had to cheat anyway, as `prepareAndInstall()` already had 4 returns :(
 Codeclimate just changed an error anyway. It doesn't like too much returns and doesn't like when I spread returns to functions :)
I'm going to postpone climate for now and test PR instead.

xetys:
report your test results then please

voron:
Manual cluster create looks good with 1 master and 1 worker : TX offload disabled on `flannel.1`, flanneld works via wg0.
But after add-worker / add-external-worker flannel ifaces are missing, I suppose due to wg0 re-creation. We need to add something like `kubectl -n kube-system delete pod -l 'app=flannel'` after `SetupEncryptedNetwork()` or maybe inside `SetupEncryptedNetwork()`

voron:
Here is corresponding [flannel issue](https://github.com/coreos/flannel/issues/877). flannel healthz check doesn't look useful too.

I added to PR and tested "delete flannel pod" work-around, now add-worker/add-external-worker perform as expected. But it's just a work-around.

voron:
@mavimo @xetys WDYT 
I don't like options like local `PostUp` wireguard hook similar to `pkill flannel`. It may be useful for those who needs to re-configure wg0 frequently, but it's out of project scope.

voron:
@xetys 
>Is it ready to be reviewed? 

Yes, the only issue is codeclimate IDK how to avoid.

>Does it work for you as expected?

Yes

itsmepetrov:
Any updates?

mavimo:
@itsmepetrov we are working to change the way we create internal network to use the Hetzner network feature.

PR #181, opened 2018-08-28
use go.mod for dependencies

JohnnyQQQQ:


xetys:
quickly for me

what's the deal of choosing go.mod over dep?

JohnnyQQQQ:
using `go mod` is the official approach introduced in golang 1.11 for handling dependencies and getting rid of the need to have the project in to GOPATH

https://github.com/golang/go/wiki/Modules

mavimo:
@JohnnyQQQQ some reference to make the CI work with Go 1.10 and 1.11 at the same time https://arslan.io/2018/08/26/using-go-modules-with-vendor-support-on-travis-ci/

xetys:
like already discussed with @mavimo I'm ok with dropping support for go <1.11, as we are not in a stable major version >= 1.0.0, and hetzner-kube more a tool than a framework.

So IMHO we can bring this PR on go 1.11 only with go mod as one package management solution.

mavimo:
@JohnnyQQQQ if needed I can work on it, let me know if you have time to complete this task :)

JohnnyQQQQ:
@mavimo feel free to pick it up, I don't have much time to work on it right now

PR #193, opened 2018-10-07
fix issue on wrong DEB packaging on k8s

mavimo:
This is a temporary fix that allow to build cluster until the issue

    https://github.com/kubernetes/kubernetes/issues/69489

will fixed


JohnnyQQQQ:
@mavimo can you add an inline comment to the issue perhaps and explain the pin in one short sentence there? So we don't forget to keep track of this.

mavimo:
@JohnnyQQQQ done

PR #199, opened 2018-10-08
Clarified Kubernetes version and host OS

rvasilev:
k8s v1.9.2 + Ubuntu 16.04

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 # hetzner-kube: fast and easy setup of kubernetes clusters on Hetzner Cloud
 
-This project contains a CLI tool to easily provision [kubernetes](https://kubernetes.io) clusters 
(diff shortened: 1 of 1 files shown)


xetys:
Here are now some conflicts left. And I don't know if it is a good idea to write the specific k8s version. As of now, it is already 1.9.6 which is installed. Later we consider adding the option to choose the version to install.

So only Ubuntu 16.04 is actually a safe statement.

mavimo:
@rvasilev the current version use k8s 1.13.2 and ubuntu 18.04 🎉

mavimo:
@rvasilev @xetys I'll suggest to close this issue once we merge https://github.com/xetys/hetzner-kube/pull/322 that report the kubernetes version managed by `hetzner-kube`, wdyt?

xetys:
Agree

PR #202, opened 2018-10-14
fix golint download path

mavimo:
SSIA


mavimo:
@JohnnyQQQQ @xetys this is needed since the old path to download golint is not longer available

mavimo:
@JohnnyQQQQ thx, I'm going to rebase other PRs :)

PR #204, opened 2018-10-14
Fix unparam processing issue and execute linting only on go 1.11.x

mavimo:
- Move linting only on stable version of golang (since they are linting do not need to be executed for each version).
 - Improve CI definition to be executed without privilege excalation
 - Define project path

Last two point become from the E2E PR, but should be moved here, that I'll rebase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: go
 
+sudo: false
+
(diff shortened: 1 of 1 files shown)


mavimo:
@JohnnyQQQQ this should be fix the issue we have on CI

PR #208, opened 2018-10-26
[WIP] Ubuntu 18.04 + k8s 1.12 + containerd 

xetys:
I started some tests to make a stable solution with Ubuntu 18.04, but it isn't really working yet.

When/if merged:
close #154


gadelkareem:
@xetys I wonder if I could help with this one. Can you explain what is exactly the problem with the upgrade?

PR #210, opened 2018-10-27
generate wireguard keypairs from go code instead of SSH in a node

mavimo:
Closes #136

PR #211, opened 2018-10-27
Add test for IP prefix & handled wrong IP

mavimo:
This should be merged after #210 

 - Refactor method `PrivateIPPrefix` that can be generate error if the source string is not a valid IP (since we are  just slicing it without any validation) 
 - Add test for method `PrivateIPPrefix`


xetys:
I see that this PR contains the changes of #210 , too. But I'm okay with both PRs

PR #212, opened 2018-10-28
allow to write file that can be accessed only from owner and use this config for wireguard

### LINKED ISSUE #165: World-readable PrivateKey in /etc/wireguard/wg0.conf
The directory `/etc/wireguard` and the file `/etc/wireguard/wg0.conf` are both world-readable, possibly exposing the `PrivateKey=` to components/users which shouldn't have access to it.

```
root@grfy-k8s-cab0-master-01 /etc/wireguard # ls -ld /etc/wireguard/
drwxr-xr-x 2 root root 4096 May  9 23:40 /etc/wireguard/
root@grfy-k8s-cab0-master-01 /etc/wireguard # ls -l /etc/wireguard/wg0.conf 
-rw-r--r-- 1 root root 989 Jul 12 01:21 /etc/wireguard/wg0.conf
```

mavimo:
Closes #165 

 - allow to write file that can be accessed only from owner
 - wireguard config can be accessed only from owner when created

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,3 +34,5 @@ require (
 	golang.org/x/text v0.3.0 // indirect
 	gopkg.in/yaml.v2 v2.2.1 // indirect
 )
+
+go 1.13
--- a/pkg/addons/addon_dashboard.go
+++ b/pkg/addons/addon_dashboard.go
@@ -66,7 +66,7 @@ subjects:
   - kind: ServiceAccount
     name: admin-user
     namespace: kube-system`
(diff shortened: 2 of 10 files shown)


mavimo:
@eliasp can you confirm that current version of code will fix the issue you reported? I tried to create a cluster and work as expected

mavimo:
@xetys rebased

xetys:
I'm so sorry...it has to be done again :/

xetys:
There was a failing build caused by the change...Maybe you find time to check what's wrong there?

mavimo:
@xetys I'll try to check ASAP

mavimo:
@xetys rebased

PR #219, opened 2018-11-04
fixed issue on CIDR parsing

mavimo:
Fixed issue introduced in #211

### DIFF
--- a/pkg/clustermanager/wireguard.go
+++ b/pkg/clustermanager/wireguard.go
@@ -47,7 +47,7 @@ Endpoint = %s:51820
 	return output
 }
 
-// PrivateIPPrefix extracts the first 3 digits of an IPv4 address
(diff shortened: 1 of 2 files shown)


pkg/clustermanager/wireguard.go
```diff
@@ -48,14 +48,14 @@ Endpoint = %s:51820
 }
 
 // PrivateIPPrefix extracts the first 3 digits of an IPv4 address
```
JohnnyQQQQ: /IPv4 address/CIDR Block

mavimo: @JohnnyQQQQ thx to take a look to this PR!
IMHO is not correct, we really extract the first 3 digits of an IPv4 address, I added that we extract if from CIDR block, let me know if it's ok.

JohnnyQQQQ: Yeah, you're right 😃

Issue #222, opened 2018-11-13
Switch to go1.11 with go.mod as dependency manager


mavimo:
SSIA

xetys:
closed via #223

PR #223, opened 2018-11-13
refs #222: move to go1.11 and go.mod

mavimo:
Signed-off-by: Marco Vito Moscaritolo <mavimo@gmail.com>


xetys:
What is about keeping the vendor dir (if there is still one left). Is this build safe?

JohnnyQQQQ:
As this is no library, I would opt-in to keep the vendor dir in the repository

mavimo:
@JohnnyQQQQ what advantages do you se to move vendor in repo? This si a bad practices from many aspect, what advantages we will have and are they enough to consider to do that?

JohnnyQQQQ:
> It's the only way to get truly reproducible builds, as it guards against upstream renames, deletes and commit history overwrites.

JohnnyQQQQ:
That's the main reason why big projects like kubernetes etc. will (always) keep the vendor dir. Otherwise they would go bonkers finding bugs sometimes.

mavimo:
@JohnnyQQQQ I can't understand:
> Otherwise they would go bonkers finding bugs sometimes.

K8s is not vendoring the codebase using go-mod since they have some blocker https://github.com/kubernetes/kubernetes/issues/68577 but is what they are try to do... same for other projects

JohnnyQQQQ:
With dep it is possible i.e. to use "master" as version reference. This is one of the main reasons to keep the vendor dir, as the master of those dependencies might evolve at any given time and break your build. 

I don't know if there is a similar problem when using gomod, but retagging can be such a problem i.e.

mavimo:
@JohnnyQQQQ `dep` allow also to use a specific commit hash that a similar approach that we can have with `go mod` that also introduce `go.sum` that include security checksum to verify that re-tagging will not introduce security issue in our codebase.

JohnnyQQQQ:
Another problem is for other devs to stay in sync. In my company we currently have a quit big monorepo written in go.

It's quit nice, that you don't need to verify your dependencies every time /so changes the dependencytree as they come with every other changes.

But this are just my 2 cents :-)

xetys:
That is mainly the reason why I also to prefer to keep the vendor directory in the repository. The packaging standards seem to change every minor version from go at the moment (go-dep, dep, go mod), as Go is quite new and things need time to become standard. 

With this conditions, it's just safer IMHO currently. I believe, there is a working solution to have everything working without keeping the vendor dir, but that looks like extra work.

mavimo:
@xetys / @JohnnyQQQQ my 2 cents:

as reference the vendoring definition is:

>>> Vendoring is the act of making your own copy of the 3rd party packages your project is using. Those copies are traditionally placed inside each project and then saved in the project repository.

Go embrace `go mod` in 1.11 as the standard approach for dependency management[1] in the go ecosystem (as npm was for node, composer for PHP, cargo for rust, ...), see [official docs](https://github.com/golang/go/wiki/Modules) where they explain that `go mod` is experimental in 1.11 and will be finalised in 1.12

Actually we do not have vendor folder committed in the `hetzner-kube` repository from [more than 3 months](https://github.com/xetys/hetzner-kube/pull/172) and I can't see issue related (maybe I'm miss some issue or maybe some contributor write to you that are not able since the vendor folder is missing from repo, but I'm unaware of it and maybe can be handled via better docs).

Commit external source code in the repository is -as far as I know- a bad practice, all language have tools that are used to handle dependency between software component (maven, gradle, composer, npm, cargo, ...) and commit 900k LoC that are C&P from other project is not a idea (there are exception, eg: company that have CI tool without access to network and they are not able to implements proxy caching, ...).

Vendoring can generate conflicts during merge that are hard to solve and allow contributors to manipulate other lib code to solve some issue, checking that the code become from vendor is the official code is really hard, and any change from official repo make impossible to upgrade project dependencies.

> Another problem is for other devs to stay in sync.

Go mod (since is integrated in the toolchain) automatically check if dependencies are correctly referenced than automatically download it at the right version, the only blocker is that go 1.11 si required.

IMHO we should avoid to reintroduce vendor folder in the project, but if all of you agree I can re-add in this PR.

--
[1] before they try to delegate this feature to external tool like dep, glide, ... but are not integrated into the go toolchain

mavimo:
@JohnnyQQQQ @xetys do you confirm that we should include all the `vendor` folder here?

Let's me know so I can close this PR :)

JohnnyQQQQ:
LGTM, how about you @xetys ?

mavimo:
@xetys up

PR #224, opened 2018-11-14
Test in nbg1-dc3 instead of fsn1-dc8

LKaemmerling:
Hello,

we saw this tweet on twitter: 

https://twitter.com/mavimo/status/1062471657825226753

Currently you try to test in fsn1-dc8, which is has not enough capacity at the moment. So I switched it to nbg1-dc3 which has enough capacity. 

For the long term, it would be better when you create the server in a location instead of a data center directly.


mavimo:
@LKaemmerling thx for the suggestion, should we use https://docs.hetzner.cloud/#datacenters-get-all-datacenters and then filter for recommended to identify the data center to use?

LKaemmerling:
@mavimo Yeah, this would be a good solution.

You can use the https://docs.hetzner.cloud/#datacenters-get-all-datacenters 
`recommendation` flag for this.

xetys:
Maybe we can do this in this PR? That would be great

mavimo:
@xetys maybe we can close this PR? @LKaemmerling we implement the E2E using the datacenter suggest via API, thx for your help!

PR #226, opened 2018-11-16
Add ufw firewall

### LINKED ISSUE #36: Firewall security
Hi, it would be nice to add some extra security and enable the firewall (ufw) in ubuntu. 

We will need some extra configuration to allow for access to Kubernetes api and other service you need to expose but easily we could add those in the provisioning script no?

gadelkareem:
Adds ufw with these rules:
```
Status: active
Logging: on (low)
Default: deny (incoming), allow (outgoing), deny (routed)
New profiles: skip

To                         Action      From
--                         ------      ----
22                         ALLOW IN    Anywhere
Anywhere                   ALLOW IN    10.0.1.0/24
Anywhere                   ALLOW IN    10.244.0.0/16
6443                       ALLOW IN    Anywhere
80                         ALLOW IN    Anywhere
443                        ALLOW IN    Anywhere
22 (v6)                    ALLOW IN    Anywhere (v6)
6443 (v6)                  ALLOW IN    Anywhere (v6)
80 (v6)                    ALLOW IN    Anywhere (v6)
443 (v6)                   ALLOW IN    Anywhere (v6)
```

fixes #36

### DIFF
--- a/pkg/clustermanager/cluster.go
+++ b/pkg/clustermanager/cluster.go
@@ -83,7 +83,7 @@ func (manager *Manager) ProvisionNodes(nodes []Node) error {
 		go func(node Node) {
 			manager.eventService.AddEvent(node.Name, "install packages")
 			//_, err := manager.nodeCommunicator.RunCmd(node, "wget -cO- https://raw.githubusercontent.com/xetys/hetzner-kube/master/install-docker-kubeadm.sh | bash -")
-			provisioner := NewNodeProvisioner(node, manager.nodeCommunicator, manager.eventService)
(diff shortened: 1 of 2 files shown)


mavimo:
@gadelkareem thx for your contribution!

I'm not sure if we should include the FW (and the define rules) in core or define it as add-on. IMHO rules that we are going to add, where we expose only few public port can prevent someone to use the config without alterate each machine manually, eg when exposed service use non 80 / 443 port, this is a good default only for web tool, but some other services should need other port that was closed (eg: if user want to expose a stateful service like MySQL, InfluxDB, Postgres, Cassandra, ...).

gadelkareem:
I agree! I will do another PR

gadelkareem:
added in #228

PR #228, opened 2018-11-19
ufw addon

### LINKED ISSUE #36: Firewall security

gadelkareem:
Add ufw to all nodes with the following rules:
```
ufw enabled with the following rules:
 Status: active
Logging: on (low)
Default: deny (incoming), allow (outgoing), deny (routed)
New profiles: skip

To                         Action      From
--                         ------      ----
Anywhere                   ALLOW IN    1.2.3.4          
Anywhere                   ALLOW IN    1.2.3.5               
22                         ALLOW IN    Anywhere                  
Anywhere                   ALLOW IN    10.0.1.0/24               
Anywhere                   ALLOW IN    10.244.0.0/16             
6443                       ALLOW IN    Anywhere                  
80                         ALLOW IN    Anywhere                  
443                        ALLOW IN    Anywhere                  
22 (v6)                    ALLOW IN    Anywhere (v6)             
6443 (v6)                  ALLOW IN    Anywhere (v6)             
80 (v6)                    ALLOW IN    Anywhere (v6)             
443 (v6)                   ALLOW IN    Anywhere (v6)  
```

fixes #36

### DIFF
--- a/pkg/addons/addon_ufw.go
+++ b/pkg/addons/addon_ufw.go
@@ -63,7 +63,6 @@ func (addon UfwAddon) Install(args ...string) {
 				" && ufw allow ssh"+
 				" && ufw allow in from "+addon.nodeCidr+" to any"+ // Kubernetes VPN overlay interface
 				" && ufw allow in from 10.244.0.0/16 to any"+ // Kubernetes pod overlay interface
-				" && ufw allow 6443"+ // Kubernetes API secure remote port
 				" && ufw allow 80"+
 				" && ufw allow 443"+
 				" && ufw default deny incoming"+


pkg/addons/addon_ufw.go
```diff
@@ -0,0 +1,89 @@
...
+				" && ufw allow in from "+addon.nodeCidr+" to any"+ // Kubernetes VPN overlay interface
+				" && ufw allow in from 10.244.0.0/16 to any"+ // Kubernetes pod overlay interface
+				" && ufw allow 6443"+ // Kubernetes API secure remote port
+				" && ufw allow 80"+
+				" && ufw allow 443"+
+				" && ufw default deny incoming"+
+				" && ufw --force enable")
+		FatalOnError(err)
+
+		output, err = addon.communicator.RunCmd(node, "ufw status verbose")
```
mavimo: Output is overridden on each iteration, maybe we should print the output (moving line 79 into foreach) or collect it here and then print?

pkg/addons/addon_ufw.go
```diff
@@ -0,0 +1,89 @@
...
+
+//Install performs all steps to install the addon
+func (addon UfwAddon) Install(args ...string) {
+
+	var nodeIpRules string
+	for _, node := range addon.nodes {
+		nodeIpRules += " && ufw allow in from " + node.IPAddress + " to any"
+	}
+	var output string
+	for _, node := range addon.nodes {
```
mavimo: We apply the same rule for each node, but I think each kind of node should have some specific rules, eg: etcd/master node don't need to expose 80/443 (that I suppose should be exposed on workers); maybe I'm missing some part of the setup can you better explain me what we like to achieve with this addon?

pkg/addons/addon_ufw.go
```diff
@@ -0,0 +1,89 @@
...
+type UfwAddon struct {
+	masterNode   *clustermanager.Node
+	communicator clustermanager.NodeCommunicator
+	nodeCidr     string
+	nodes        []clustermanager.Node
+}
+
+//NewUfwAddon installs ufw to the cluster
+func NewUfwAddon(provider clustermanager.ClusterProvider, communicator clustermanager.NodeCommunicator) ClusterAddon {
+	masterNode, _ := provider.GetMasterNode()
```
mavimo: Unhandled error, we prefer to never ignore errors ;)

pkg/addons/addon_ufw.go
```diff
@@ -0,0 +1,89 @@
...
+	return UfwAddon{masterNode: masterNode, communicator: communicator, nodeCidr: provider.GetNodeCidr(), nodes: provider.GetAllNodes()}
```
mavimo: IMHO move in a multiline should make the code more readable, like:
```
return UfwAddon{
    masterNode: masterNode,
    communicator: communicator,
    nodeCidr: provider.GetNodeCidr(),
    nodes: provider.GetAllNodes(),
}
```

pkg/addons/addon_ufw.go
gadelkareem: I am not sure if we want to print the output for each server, especially when it is fairly the same output. Let me know your thoughts?

mavimo: Actually we are going to execute a command than report the result of the command execution to user as "changes applied to all nodes".

IMHO this is not correct because (as reported above) we can have different rules based on node type, and should be "correct" to explain the change we add to each node or just report the info on expected reported situation to the user (that do not involve UFW execution report on a single node but the rule we are going to apply).

Let's suppose the last node of the iteration already have some rule define (maybe manually?) that are not configured on all nodes, a user can suppose that the result (already existing rule + rule from addon) are configured on all nodes but this is not true.

Maybe is an edge case and we prefer to ignore this situation?

pkg/addons/addon_ufw.go
gadelkareem: So I think the ufw rules are the trickiest part of this addon. I tried to use the most common rules any new cluster would require. But you can always change these rules via script runner from #229 for ex.

xetys:
whats the state here?

willipreuk:
Any updates?

tckb:
This looks pretty important. Whats the state of this PR?

xetys:
push @mavimo @gadelkareem

mavimo:
@xetys I'll take a look in the next few days..

liebig:
Any news on this?

alexanderkjeldaas:
Any update?  This was referred to by https://www.tmjohnson.co.uk/posts/k8s/

stevek-pro:
Any update here? We need this 2 years ago.

PR #229, opened 2018-11-19
Script Runner addon

gadelkareem:
Script Runner makes it easy to run a bash script on all your k8s nodes with one command.
```
hetzner-kube cluster addon install --name demo script-runner /path/to/script.sh 
```

### DIFF
--- a/cmd/cluster_addon.go
+++ b/cmd/cluster_addon.go
@@ -51,7 +51,7 @@ func validateAddonSubCommand(cmd *cobra.Command, args []string) error {
 	if idx == -1 {
 		return fmt.Errorf("cluster '%s' not found", name)
 	}
-	if len(args) != 1 {
+	if len(args) < 1 {
 		return errors.New("exactly one argument expected")
 	}
 	addonName := args[0]
--- a/pkg/addons/addon_script_runner.go
+++ b/pkg/addons/addon_script_runner.go
@@ -1,0 +1,75 @@
+package addons
+
+import (
+	"fmt"
+
+	"github.com/xetys/hetzner-kube/pkg/clustermanager"
+	"io/ioutil"
+	"log"
+	"time"
+)
+
+//ScriptRunnerAddon installs script runner
+type ScriptRunnerAddon struct {
+	masterNode   *clustermanager.Node
+	communicator clustermanager.NodeCommunicator
+	nodes        []clustermanager.Node
+}
+
+//NewScriptRunnerAddon installs script runner to the cluster
+func NewScriptRunnerAddon(provider clustermanager.ClusterProvider, communicator clustermanager.NodeCommunicator) ClusterAddon {
+	masterNode, _ := provider.GetMasterNode()
+	return ScriptRunnerAddon{masterNode: masterNode, communicator: communicator, nodes: provider.GetAllNodes()}
+}
+
+func init() {
+	addAddon(NewScriptRunnerAddon)
+}
+
+//Name returns the addons name
+func (addon ScriptRunnerAddon) Name() string {
+	return "script-runner"
+}
+
+//Requires returns a slice with the name of required addons
+func (addon ScriptRunnerAddon) Requires() []string {
+	return []string{}
+}
+
+//Description returns the addons description
+func (addon ScriptRunnerAddon) Description() string {
+	return "Bash remote script runner"
+}
+
+//URL returns the URL of the addons underlying project
+func (addon ScriptRunnerAddon) URL() string {
+	return "https://www.gnu.org/software/bash/"
+}
+
+//Install performs all steps to install the addon
+func (addon ScriptRunnerAddon) Install(args ...string) {
+
+	if len(args) < 1 {
+		log.Fatal("path argument is missing")
+	}
+	scriptPath := args[1]
+	scriptContents, err := ioutil.ReadFile(scriptPath)
+	FatalOnError(err)
+
+	var output string
+	for _, node := range addon.nodes {
+		scriptRemotePath := "/tmp/script-" + time.Now().Format("20060102150405") + ".sh"
+		err = addon.communicator.WriteFile(node, scriptRemotePath, string(scriptContents), true)
+		FatalOnError(err)
+
+		output, err = addon.communicator.RunCmd(node, "bash "+scriptRemotePath)
+		FatalOnError(err)
+	}
+
+	fmt.Println("Script ran successfully:\n", output)
+}
+
+//Uninstall performs all steps to remove the addon
+func (addon ScriptRunnerAddon) Uninstall() {
+	fmt.Println("no uninstall for this addon")
+}
--- a/cmd/cluster_addon_install.go
+++ b/cmd/cluster_addon_install.go
@@ -44,7 +44,7 @@ var clusterAddonInstallCmd = &cobra.Command{
 		FatalOnError(err)
 
 		addon := addonService.GetAddon(addonName)
-		addon.Install()
+		addon.Install(args...)
 
 		log.Printf("addon %s successfully installed", addonName)
 	},


xetys:
wow, that's I cool. I'll review asap

mavimo:
Yeah! I suggest to allow to specify the server type, actually we have 3 main "type" of server, `etcd`, `master` and `worker`.. I think most of time we need to run specific script only on a specific type of server, so maybe have a flag to define what kind of server should be affected from script execution should be a big improvement.

gadelkareem:
@mavimo we can use the power of bash and keep the addon simple :)
```
#!/bin/bash

if [[ $(hostname -s) = *master* ]]; then
    exit 0
fi
```

mavimo:
@gadelkareem it's a possibility, so the addon is simple but bash script should be "complex" ;)

I think is also a bit fragile because we can't guarantee that the name of server match the *master* / *worker* / *etcd* name, we generate new server with the specified name but a user can add the server to the pool with a different name or rename it.. so IMHO we should move server type definition on addon instead of into the script that we deploy

gadelkareem:
@mavimo I see that the names are not constant, especially if it is an external server being added. Maybe we should have a file on the server to hint its type?

I think the bash script will probably be much more complex than the hostname check example above. And since the user already knows his own cluster then it should be trivial to manage the script based on the hostname or any other properties. Because if we start adding a new argument for each use case then it will not be as simple as now.

For example, with the current use case of choosing the instance group to run on, we will need to think if the user also wants to run the script on 2 groups (master and worker) or 3 of them. And what if there is a new group created then we need to modify this addon. That's why I always prefer to keep it simple and if we have feature request then it could be added to it.

mavimo:
@gadelkareem WDYT to send as args to script invocation some information about current server / cluster topology?

gadelkareem:
@mavimo good idea! which info do you think would be nice to pass to the script?

mavimo:
@gadelkareem NB: I'm not sure but what we should pass, I'll try to write some proposal for discussion 😄 

We should start with something like:

 - **Server type** (master, worker, etcd) or a combination of them (some server should have more than one role at time) in order to allow script to have specific config (eg port exposed by firewall? ;) )
 - **Master IP**:  because some time script should need to dialog with the K8s API for specific operation
 - **Cluster type** (without HA, with HA level 3, with HA level 4, ..) or **List of server configs** in order to make easy to add action that dialogue with other servers in pool

Maybe point 2 and 3 should be mixed if we build a JSON that contains all cluster information as second argument like:

     /tmp/script-to-run.sh master {"masters":[{"name":"master-1","ip":"111.222.0.1"}]...}

where script should process the second information if needed (eg: `echo $2 | jq .masters[].ip`)

gadelkareem:
I also allowed the display of output for each node since it will be different.

xetys:
The LGTM already, but I would love to see some docs for the users on how to use that. With some examples, so users will know how to deal with the params. Maybe a distinct page like the HA-guide and wireguard info page?

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"encoding/json"
+	"github.com/xetys/hetzner-kube/pkg/clustermanager"
+	"io/ioutil"
+	"log"
+	"strings"
+	"time"
+)
+
+//ScriptRunnerAddon installs script runner
```
mavimo: Single space between `//` and the next part of comment (on all line)

```suggestion
// ScriptRunnerAddon installs script runner
```

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+type ScriptRunnerAddon struct {
+	communicator clustermanager.NodeCommunicator
+	nodes        []clustermanager.Node
+	cluster      clustermanager.Cluster
+}
+
+//NewScriptRunnerAddon installs script runner to the cluster
```
mavimo:
```suggestion
// NewScriptRunnerAddon installs script runner to the cluster
```

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+func NewScriptRunnerAddon(provider clustermanager.ClusterProvider, communicator clustermanager.NodeCommunicator) ClusterAddon {
+	return ScriptRunnerAddon{communicator: communicator, nodes: provider.GetAllNodes(), cluster: provider.GetCluster()}
+}
+
+func init() {
+	addAddon(NewScriptRunnerAddon)
+}
+
+//Name returns the addons name
```
mavimo:
```suggestion
// Name returns the addons name
```

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+func (addon ScriptRunnerAddon) Name() string {
+	return "script-runner"
+}
+
+//Requires returns a slice with the name of required addons
```
mavimo:
```suggestion
// Requires returns a slice with the name of required addons
```

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+func (addon ScriptRunnerAddon) Requires() []string {
+	return []string{}
+}
+
+//Description returns the addons description
```
mavimo:
```suggestion
// Description returns the addons description
```

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+func (addon ScriptRunnerAddon) Description() string {
+	return "Bash remote script runner"
+}
+
+//URL returns the URL of the addons underlying project
```
mavimo:
```suggestion
// URL returns the URL of the addons underlying project
```

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+func (addon ScriptRunnerAddon) URL() string {
+	return "https://www.gnu.org/software/bash/"
+}
+
+//Install performs all steps to install the addon
```
mavimo:
```suggestion
// Install performs all steps to install the addon
```

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+			"bash "+
+				scriptRemotePath+
+				" "+node.Group+
+				" '"+clusterInfo+"'")
+		FatalOnError(err)
+		fmt.Println(node.Name+" "+node.IPAddress+": script ran successfully..\n", output)
+	}
+}
+
+//Uninstall performs all steps to remove the addon
```
mavimo:
```suggestion
// Uninstall performs all steps to remove the addon
```

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	scriptPath := args[1]
+	scriptContents, err := ioutil.ReadFile(scriptPath)
+	FatalOnError(err)
+
+	clusterInfoBin, err := json.Marshal(addon.cluster)
+	FatalOnError(err)
+
+	replacer := strings.NewReplacer("\n", "", "'", "\\'")
+	clusterInfo := replacer.Replace(string(clusterInfoBin))
```
mavimo: I'm not sure to understand why we are going to use replacer here, this is needed since there are some info in JSON encoded that can generate invalid string argument? Is possibile to use some different approach in order to do not manipulate result value, like:

    echo $(cat <<EOF
    <JSON_MARSHALLED_VALUE>
    EOF)

I'm not sure is doable, but will allow us to do not manipulate the value and be "safe" on changes we do in config..

pkg/addons/addon_script_runner.go
gadelkareem: Although I added the suggested changes. I think this is personal preference, we should stick to what "go fmt" gives us to avoid wasting time on these trivial changes.

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
+package addons
+
+import (
+	"fmt"
```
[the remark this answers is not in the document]
gadelkareem: hmm.. did you mean remove empty line?

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+	FatalOnError(err)
+
+	clusterInfoBin, err := json.Marshal(addon.cluster)
+	FatalOnError(err)
+
+	replacer := strings.NewReplacer("\n", "", "'", "\\'")
+	clusterInfo := replacer.Replace(string(clusterInfoBin))
+
+	for _, node := range addon.nodes {
+		scriptRemotePath := "/tmp/script-" + time.Now().Format("20060102150405") + ".sh"
```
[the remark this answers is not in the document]
gadelkareem: You do not need a random number here, since the time is formatted with milliseconds.

pkg/addons/addon_script_runner.go
mavimo: @gadelkareem this is the suggested format for go stdlib like https://github.com/golang/go/blob/master/src/os/env.go

I'm not sure `go fmt` will process the comments, there are some issue related to that but the core team decide that comments should not be manipulated from fmt, see https://github.com/golang/go/issues/5128

pkg/addons/addon_script_runner.go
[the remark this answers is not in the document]
mavimo: I mean that we should change it to:
```
import (
	"encoding/json"
	"fmt"
	"io/ioutil"
	"log"
	"strings"
	"time"

	"github.com/xetys/hetzner-kube/pkg/clustermanager"
)
```

pkg/addons/addon_script_runner.go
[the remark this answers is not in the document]
mavimo: Actually we are going to print only seconds:
```go
package main

import (
	"fmt"
	"time"
)

func main() {
	fmt.Println(time.Now().Format("20060102150405"))
	time.Sleep(10 * time.Microsecond)
	fmt.Println(time.Now().Format("20060102150405"))
	time.Sleep(10 * time.Microsecond)
	fmt.Println(time.Now().Format("20060102150405"))
}
```
and IMHO this can generate some clash

pkg/addons/addon_script_runner.go
```diff
@@ -0,0 +1,85 @@
...
+		err = addon.communicator.WriteFile(node, scriptRemotePath, string(scriptContents), true)
+		FatalOnError(err)
+
+		output, err := addon.communicator.RunCmd(
+			node,
+			"bash "+
```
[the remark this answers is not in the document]
gadelkareem: the %q won't work because of it uses double-quoted

pkg/addons/addon_script_runner.go
gadelkareem: I don't think the run command will allow multi line. If it works for you, please let me know.

pkg/addons/addon_script_runner.go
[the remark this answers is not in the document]
gadelkareem: yeah updated

cmd/cluster_addon.go
```diff
@@ -51,8 +51,8 @@ func validateAddonSubCommand(cmd *cobra.Command, args []string) error {
 	if idx == -1 {
 		return fmt.Errorf("cluster '%s' not found", name)
 	}
-	if len(args) != 1 {
-		return errors.New("exactly one argument expected")
+	if len(args) < 1 {
```
mavimo: One question, since is not clear to me the expected behaviour.

Can I invoke this add-on with more than one argument and the first one should be the script, eg:
```
hetzner-kube cluster addon install --name script-runner /PATH/TO/MY/SCRIPT FIRST_ARG SECOND_ARG THIRD_ARG
```
or I need to run it with only one argument that indicate the path script file like:
```
hetzner-kube cluster addon install --name script-runner /PATH/TO/MY/SCRIPT
```
In the first case I expect that extra args are propagated to the script on the server, in the second case I expect to have an error if there are some extra args that will be ignored.

gadelkareem: No, extra args will not be sent to the script. The correct call should be 
```
hetzner-kube cluster addon install --name script-runner /PATH/TO/MY/SCRIPT
```
And the script will be called with node group and cluster info as args, ex:
```
/tmp/script.sh worker '{...}'
```

mavimo: Than maybe is better that we check that `if len(args) != 2 { }` and if not return an error.

gadelkareem: I think since this validation is for all addons, we do not want to restrict it.

mavimo: Right, maybe we should add a new method `ArgsValidation` to the `Addon` interface and delegate each plugin to validate data they need (I mean in a different PR), WDYT?

gadelkareem: IMHO I would leave it up to the addon to decide it without any restriction.

mavimo:
@gadelkareem do you have time to write some doc? If you need some help please ping me :)

gadelkareem:
@mavimo Since you are asking. Do you have an example config.json that i can use?

gadelkareem:
added some docs!

docs/script-runner-addon.md
```diff
@@ -0,0 +1,28 @@
+# Script Runner Addon Guide
+
+Script Runner lets you run a bash script on all your cluster nodes with one command. It uploads the bash script to each node then execute. The script will receive two arguments upon execution:
+1) Node group, for example (master, worker, etcd)
```
mavimo: Markdown suggest to use:

```suggestion

1. Node group, allowed values are `master`, `worker` and `etcd`
```

docs/script-runner-addon.md
```diff
@@ -0,0 +1,28 @@
+# Script Runner Addon Guide
+
+Script Runner lets you run a bash script on all your cluster nodes with one command. It uploads the bash script to each node then execute. The script will receive two arguments upon execution:
+1) Node group, for example (master, worker, etcd)
+2) Cluster info JSON config, which is similar to the hetzner-kube config found in `~/.hetzner-kube/config.json`
```
mavimo: Markdown suggest to use
```suggestion
1. Cluster info JSON config, which is similar to the `hetzner-kube` configuration that you can  found in `~/.hetzner-kube/config.json`
```

docs/script-runner-addon.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Example
+hetzner-kube Command:
+```bash
+hetzner-kube cluster addon install --name demo script-runner /path/to/script.sh  
+```
+Example script.sh:
+```bash
+
```
mavimo: Bash script should start with:
```
#!/bin/bash
```
this should be omitted (than specify on the doc that should not be included) or is missing here (than add to the sample)?

docs/script-runner-addon.md
```diff
@@ -0,0 +1,28 @@
+# Script Runner Addon Guide
+
+Script Runner lets you run a bash script on all your cluster nodes with one command. It uploads the bash script to each node then execute. The script will receive two arguments upon execution:
+1) Node group, for example (master, worker, etcd)
+2) Cluster info JSON config, which is similar to the hetzner-kube config found in `~/.hetzner-kube/config.json`
+
+### Example
+hetzner-kube Command:
```
mavimo: Please add a space between title and text and between text and code sample

mavimo:
@gadelkareem I know I should sound like a "nuisance" (sorry), added some comments to the doc :)

gadelkareem:
cool done ;)

PR #230, opened 2018-11-20
Add the optional k8s-version flag to cobra

### LINKED ISSUE #162: kubernetes version select
I see hetzner-kube supports k8s version select via [undocumented flag](https://github.com/xetys/hetzner-kube/blob/5ebf7a149be2c6066525e6ef0cb403e66ed840a8/pkg/clustermanager/provision_node.go#L13). Is there any option instead of patching default value in source and re-build hetzner-kube to choose k8s version to be deployed?

gadelkareem:
fixes #162

### DIFF
--- a/cmd/cluster_create.go
+++ b/cmd/cluster_create.go
@@ -308,6 +308,7 @@ func init() {
 	clusterCreateCmd.Flags().StringP("ssh-key", "k", "", "Name of the SSH key used for provisioning")
 	clusterCreateCmd.Flags().String("master-server-type", "cx11", "Server type used of masters")
 	clusterCreateCmd.Flags().String("worker-server-type", "cx11", "Server type used of workers")
+	clusterCreateCmd.Flags().String("k8s-version", "1.9.6-00","The version of the k8s debian packages that will be used during provisioning")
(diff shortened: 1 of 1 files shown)


mavimo:
IMHO should be a nice starting point, I think some more changes are needed:

 - Update README to allow to specify what value are accepted, until now I think the cluster should be accept 1.9.X, 1.10.X and 1.11.X since config format is unchanged in this versions (as far I remember), but we should do some more checks
 - E2E test should be added to allow to deploy a cluster for each supported version (maybe a small cluster for each version should be fine)
 - Config should store the cluster version and it should be restored on load in order to apply consistent behaviour when needed

IMHO this should be a breaking change, so we should dump version and maybe we can consider to implement it in 0.4 where @xetys is considering to allow to have support to the K8s 1.12 (see #208), but I'm not sure we will support multiple cluster version in the application at the first iteration.

gadelkareem:
@mavimo and thanks for great project ;)

I actually did test it but unfortunately all versions failed.  That's why it is a good idea to leave it out of the docs for now and merge it for new upcoming versions.

PR #231, opened 2018-11-22
[WIP] Feat ubuntu 18

gadelkareem:
I switched to Weave Net and everything seems to work!

I tried to do the PR against feat-ubuntu-18 from #208 but github doesn't seem to show to correct changes.

Still getting an error on the join command although the node is being added.

### DIFF
--- a/pkg/clustermanager/cluster.go
+++ b/pkg/clustermanager/cluster.go
@@ -151,8 +151,10 @@ func (manager *Manager) InstallMasters() error {
 	commands := []NodeCommand{
 		{"kubeadm init", "kubeadm init --config /root/master-config.yaml"},
 		{"configure kubectl", "rm -rf $HOME/.kube && mkdir -p $HOME/.kube && cp -i /etc/kubernetes/admin.conf $HOME/.kube/config && chown $(id -u):$(id -g) $HOME/.kube/config"},
-		{"install flannel", "kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/v0.10.0/Documentation/kube-flannel.yml"},
-		{"configure flannel", "kubectl -n kube-system patch ds kube-flannel-ds --type json -p '[{\"op\":\"add\",\"path\":\"/spec/template/spec/tolerations/-\",\"value\":{\"key\":\"node.cloudprovider.kubernetes.io/uninitialized\",\"value\":\"true\",\"effect\":\"NoSchedule\"}}]'"},
+		{"install canal (RBAC)", "kubectl apply -f https://docs.projectcalico.org/v3.1/getting-started/kubernetes/installation/hosted/canal/rbac.yaml"},
+		{"install canal", "kubectl apply -f https://docs.projectcalico.org/v3.1/getting-started/kubernetes/installation/hosted/canal/canal.yaml"},
+		//{"install flannel", "kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/v0.10.0/Documentation/kube-flannel.yml"},
+		//{"configure flannel", "kubectl -n kube-system patch ds kube-flannel-ds --type json -p '[{\"op\":\"add\",\"path\":\"/spec/template/spec/tolerations/-\",\"value\":{\"key\":\"node.cloudprovider.kubernetes.io/uninitialized\",\"value\":\"true\",\"effect\":\"NoSchedule\"}}]'"},
 		//{"install hcloud integration", fmt.Sprintf("kubectl -n kube-system create secret generic hcloud --from-literal=token=%s", AppConf.CurrentContext.Token)},
 		//{"deploy cloud controller manager", "kubectl apply -f  https://raw.githubusercontent.com/hetznercloud/hcloud-cloud-controller-manager/master/deploy/v1.0.0.yaml"},
 	}
--- a/pkg/clustermanager/cluster.go
+++ b/pkg/clustermanager/cluster.go
@@ -169,14 +169,14 @@ func (manager *Manager) InstallMasters() error {
 
 	for _, node := range manager.nodes {
 		if node.IsMaster {
-			_, err := manager.nodeCommunicator.RunCmd(node, "kubeadm reset")
+			_, err := manager.nodeCommunicator.RunCmd(node, "kubeadm reset -f")
 			if err != nil {
-				return nil
+				return err
 			}
 
 			_, err = manager.nodeCommunicator.RunCmd(node, "rm -rf /etc/kubernetes/pki && mkdir /etc/kubernetes/pki")
 			if err != nil {
-				return nil
+				return err
 			}
 			if len(manager.nodes) == 1 {
 				commands = append(commands, NodeCommand{"taint master", "kubectl taint nodes --all node-role.kubernetes.io/master-"})
@@ -325,6 +325,7 @@ func (manager *Manager) InstallWorkers(nodes []Node) error {
 	if err != nil {
 		return err
 	}
+	joinCommand = fmt.Sprintf("%s --cri-socket /var/run/docker/containerd/docker-containerd.sock", joinCommand)
 
 	errChan := make(chan error)
 	trueChan := make(chan bool)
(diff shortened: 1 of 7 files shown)


mavimo:
@gadelkareem should a good starting point, thx!

I'll try to review the changes this weekend (sorry, need some time and I'm a bit busy, just few note:

 - I suggest to move the debug flag to the root command and than propagate to the other commands (or maybe move this feature in a different PR)
 - The `dnsDomain` is hardcoded to `cluster.local`, I'm not sure it should be configurable or not
 - Need some test ;)

Thanks again for the effort you place on improve the original PR!

pkg/clustermanager/provision_node.go
```diff
@@ -139,7 +152,7 @@ func (provisioner *NodeProvisioner) prepareDocker() error {
 	// docker-ce
 	aptPreferencesDocker := `
 Package: docker-ce
-Pin: version 17.03.*
+Pin: version 18.06.0~ce~3-0~ubuntu
```
xetys: I had issues with this one. It still installed 18.06.1 in my case. Did that change to you? 

Anyway, I'm not sure if this approach is actually a good practice. It's the first and only thing which actually worked

pkg/clustermanager/wireguard.go
```diff
@@ -82,3 +82,23 @@ func GenerateKeyPair() (WgKeyPair, error) {
...
+func GenerateOverlayRouteSystemdService(node Node) string {
+	serviceTpls := `# /etc/systemd/system/overlay-route.service
+[Unit]
+Description=Overlay network route for Wireguard
+After=wg-quick@wg0.service
+
+[Service]
+Type=oneshot
+User=root
+ExecStart=/sbin/ip route add 10.96.0.0/12 dev wg0 src %s
```
xetys: is that for routing pod networking over wireguard? Did you check that

pkg/clustermanager/cluster.go
```diff
@@ -152,10 +160,13 @@ func (manager *Manager) SetupEncryptedNetwork() error {
 func (manager *Manager) InstallMasters() error {
 
 	commands := []NodeCommand{
-		{"kubeadm init", "kubeadm init --config /root/master-config.yaml"},
+		{"kubeadm init", "kubectl version > /dev/null &> /dev/null || kubeadm init --config /root/master-config.yaml"},
 		{"configure kubectl", "rm -rf $HOME/.kube && mkdir -p $HOME/.kube && cp -i /etc/kubernetes/admin.conf $HOME/.kube/config && chown $(id -u):$(id -g) $HOME/.kube/config"},
-		{"install flannel", "kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/v0.10.0/Documentation/kube-flannel.yml"},
-		{"configure flannel", "kubectl -n kube-system patch ds kube-flannel-ds --type json -p '[{\"op\":\"add\",\"path\":\"/spec/template/spec/tolerations/-\",\"value\":{\"key\":\"node.cloudprovider.kubernetes.io/uninitialized\",\"value\":\"true\",\"effect\":\"NoSchedule\"}}]'"},
+		{"install Weave Net", "kubectl apply -f \"https://cloud.weave.works/k8s/net?k8s-version=$(kubectl version | base64 | tr -d '\\n')\""},
```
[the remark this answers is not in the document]
gadelkareem: I cannot confirm for sure if there was a problem with canal because I switched to Weave Net following few articles which were recommending it for its easy no config setup. I am not sure what is the advantage of canal though..

pkg/clustermanager/cluster.go
```diff
@@ -338,7 +350,10 @@ func (manager *Manager) InstallWorkers(nodes []Node) error {
 			numProcs++
 			go func(node Node) {
 				manager.eventService.AddEvent(node.Name, "registering node")
-				_, err := manager.nodeCommunicator.RunCmd(node, "kubeadm reset && "+joinCommand)
+				_, err := manager.nodeCommunicator.RunCmd(
+					node,
+					"for i in ip_vs ip_vs_rr ip_vs_wrr ip_vs_sh nf_conntrack_ipv4; do modprobe $i; done"+
```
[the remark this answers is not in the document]
gadelkareem: This could be the main issue why kubeadm join command was failing which is related to https://github.com/kubernetes/kubeadm/issues/975

So enabling those modules fixed the warning.

pkg/clustermanager/provision_node.go
gadelkareem: I believe it worked for me. There was no problems with the containers, I also tested with rook.io storage and everything worked fine. 

Which problems did you face with this one?

pkg/clustermanager/wireguard.go
gadelkareem: That one is for Weave Net to make sure it routes traffic via the correct network interface. More on it here https://github.com/hobby-kube/guide

gadelkareem:
> @gadelkareem should a good starting point, thx!
> 
> I'll try to review the changes this weekend (sorry, need some time and I'm a bit busy, just few note:
> 
> * I suggest to move the debug flag to the root command and than propagate to the other commands (or maybe move this feature in a different PR)
> * The `dnsDomain` is hardcoded to `cluster.local`, I'm not sure it should be configurable or not
> * Need some test ;)
> 
> Thanks again for the effort you place on improve the original PR!

Sure tyt..

* The debug log was very important to see all the ssh commands. So maybe we can split it out when this PR is ready.
* I don't think the dnsDomain means much to the setup.
* Yep! just let me know your findings..

pkg/clustermanager/cluster.go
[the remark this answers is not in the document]
xetys: it allows to enforce Pod Security Policies as of using callico. It is actually the same simple to setup as Weave or Flannel....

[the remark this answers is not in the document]
gadelkareem: I just tested it and it works fine. Even without the ip routing command. I am not sure which we should stick to.

pkg/clustermanager/cluster.go
```diff
@@ -342,9 +341,6 @@ func (manager *Manager) InstallWorkers(nodes []Node) error {
 	}
 	joinCommand = fmt.Sprintf("%s --apiserver-advertise-address=10.0.0.1 --cri-socket /var/run/docker/containerd/docker-containerd.sock", joinCommand)
 
-	log.Println("sleep for 10s...")
```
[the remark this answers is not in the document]
xetys: what's about this one?

pkg/clustermanager/cluster.go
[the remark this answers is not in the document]
xetys: I would prefer canal over Weave, if that works

pkg/clustermanager/provision_node.go
xetys: It failed during kubeadm test phase to talk to the container runtime, as it didn't work to communicate. It is a known issue for 18.06.1. That actually makes me super-skeptic, as I'm not convinced if that approach is the right one. It's still the only working solution to make k8s run on Ubuntu 18 with a supported runtime.

Also interesting if 18.09 is already available and if this approach is working with it.

Could you please check that?

pkg/clustermanager/wireguard.go
xetys: If we do this for canal, maybe we should check for the correct IP range.

pkg/clustermanager/cluster.go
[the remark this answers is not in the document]
gadelkareem: It is outdated. I think I added at some point then removed it.

pkg/clustermanager/cluster.go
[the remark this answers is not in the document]
gadelkareem: I switched back to it. Although I see weave is much more popular..

pkg/clustermanager/provision_node.go
gadelkareem: 18.09 has a problem with crictl binary not understand the new API. But 18.06 works fine.

pkg/clustermanager/wireguard.go
gadelkareem: added!

pushrbx:
I read through the code as I was planning to use this PR to setup my cluster, and I found one thing which I think should be addressed in this PR or in another:
https://github.com/gadelkareem/hetzner-kube/blob/3b81629d7edf342008085e62aa170e09fbe67e48/pkg/addons/addon_hcloud_controller_manager.go#L69
Some addons are still using flannel related commands.

pkg/clustermanager/provision_node.go
xetys: Just to review, with this PR you definitely get 18.06.0 and not 18.06.1, which is not working?

If yes, I'm OK with it

gadelkareem: I am not sure about 18.06.1, but 18.06.0 is working and 18.09 is not, according to my tests.

PR #243, opened 2019-01-27
Ubuntu 18.04 & Kubernetes 1.13.2 & containerd

mavimo:
SSIA

Based on work from @xetys and @gadelkareem


mavimo:
@xetys I'm not sure we need to fix the codeclimate issue, the template needed from test should not be reduced and move it outside test is not a good idea IMO

PR #250, opened 2019-02-19
Add Kubernetes Dashboard addon

### LINKED ISSUE #83: can we also add kubernetes-dashboard as addon
I think this would be of great value, to get a quick overview over the newly created cluster.

cornelk:
Closes #83

### DIFF
--- a/pkg/addons/addon_dashboard.go
+++ b/pkg/addons/addon_dashboard.go
@@ -85,5 +85,8 @@ func (addon DashboardAddon) Uninstall() {
 	_, err := addon.communicator.RunCmd(node, "kubectl delete -f https://raw.githubusercontent.com/kubernetes/dashboard/v1.10.1/src/deploy/recommended/kubernetes-dashboard.yaml")
 	FatalOnError(err)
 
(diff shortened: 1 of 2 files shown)


pkg/addons/addon_dashboard.go
```diff
@@ -0,0 +1,89 @@
...
+	return "https://github.com/kubernetes/dashboard"
+}
+
+// Install performs all steps to install the addon
+func (addon DashboardAddon) Install(args ...string) {
+	node := *addon.masterNode
+	_, err := addon.communicator.RunCmd(node, "kubectl apply -f https://raw.githubusercontent.com/kubernetes/dashboard/v1.10.1/src/deploy/recommended/kubernetes-dashboard.yaml")
+	FatalOnError(err)
+
+	serviceAccount := `apiVersion: v1
```
xetys: you create a new file here. Do you clean up the service account in uninstall? If not, please add it so this addon doesn't leave stuff in the cluster plz

cornelk: fixed

Issue #234, opened 2018-11-25
Developer experience improvement (DX initiative)


mavimo:
Hi guys,
there are some "bug" on user experience and I'll like to fix it, but before start I think we should find a solution that should match the expectation of much people as possibile.

### Command consistency

As end-user there are command that are not consistent on usage, some time we use flag, some time we need to user args, eg:

 - `hetzner-cloud context delete` use **arg** do indicate the name of context to delete
 - `hetzner-cloud ssh-key delete` use **flag** do indicate the name of ssh-key to delete
 - `hetzner-cloud cluster delete` use **flag** do indicate the name of cluster to delete

Same for 
 - `hetzner-cloud context add` use **arg** do indicate the name of context to add
 - `hetzner-cloud ssh-key add` use **flag** do indicate the name of ssh-key to add
 - `hetzner-cloud cluster add` use **flag** do indicate the name of cluster to add

etc.

Maybe we should define the standard behaviour we will like to use and refactor other commands in order to be consistent.

### Debug mode

There is a `--debug` flag defined on the root command but other command do not honor it, maybe we should allow all command to use this information when required.

### Command listing

The `hetzner-cloud` list only the first level command, so user should enter in each command to detect the list of available subcommand, maybe we should list all command we have as doc? Something like:

```
A tool for creating and managing kubernetes clusters on Hetzner Cloud.

Usage:
  hetzner-kube [command]

Available Commands:
  cluster     view and manage kubernetes clusters, include subcommands:
     - add-external-worker    adds an existing server to the cluster
     - add-worker             add worker nodes
     - addon                  manages addons for kubernetes clusters
     - create                 creates a cluster
     - delete                 removes a cluster and deletes the associated nodes
     - kubeconfig             setups the kubeconfig for the local machine
     - list                   lists all created clusters
     - master-ip              get master node ip
     - remove-external-worker remove an external worker from node list

  completion  Generates bash completion scripts

  context     view and manage contexts, include subcommands:
     - add         adds a new context
     - current     prints the current used context
     - delete      deletes a new context
     - list        lists all contexts
     - use         switches to a saved context given by NAME

  help        Help about any command

  ssh-key     view and manage SSH keys, include subcommands:
     - add         adds a new SSH key to the Hetzner Cloud project and local configuration
     - delete      removes a saved SSH key from local configuration and Hetzner Cloud account
     - list        lists all saved SSH keys

  version     prints the current version

Flags:
      --config string   config file to use
[...]

PR #255, opened 2019-02-28
Kubeconfig command features update

md2k:
This PR about `hetzner-kube cluster kubeconfig my-cluster`
What added:
- [x] config normalization (rewrite default context/cluster/authinfo names to current cluster-name)
- [x] by default if nothing is set, just print current config for selected cluster
- [x] save to specific location, by default it saves to `~/.kube/my-cluster.yaml`
- [ ] merge - merges current and picked cluster's config


md2k:
@xetys  @mavimo  can you take a look.
`merge` functionality still in progress, but near weekend will be done.
also will be nice if you can point me why go.mod have merge conflicts if i had latest from master before any changes.  i'm bit old-school guy and not yet widely use go mod... prefer old way vendor-ing :)

mavimo:
@md2k nice! I'm also working on refactoring the command structure using item as discussed on https://github.com/xetys/hetzner-kube/issues/234

md2k:
To perform merge i'm going to use same approach as for sanitization of config (client-go, they have all required api for this), stole this way from official kubectl part, just a bit light weight for our needs. before EOW planing to add merge here.

md2k:
@mavimo @xetys  added merge, tested multiple time with my local setup. it is safe to use.
Current behavior simply overwrite if same context already exists. not sure if we need to add `force` switch, because if i do merge for same cluster-name, probably i did complete rebuild of cluster... or .. do we need verification if i really want overwrite context?

mavimo:
@md2k JM2C: we should not overwrite a context, we should communicate the failure to the user that can rename/delete the context before retry to pull it.

Adding a flag `--force` should be a nice enhancement but not strictly needed from my PoV.

cmd/util.go
```diff
@@ -17,3 +19,29 @@ func FatalOnError(err error) {
...
+
+// GetHome is an helper function to get current home directory
+func GetHome() string {
+	usr, _ := user.Current()
+	return usr.HomeDir
+}
+
+// Dump is handy dumps of structure as json (almost any structures)
+func Dump(cls interface{}) {
+	data, err := json.MarshalIndent(cls, "", "    ")
```
mavimo: Maybe we can replace it with 
```
fmt.Println(Sdump(cli))
```
Since I suppose this is a debug utils, I'm not sure we should keep it in our codebase :)

cmd/cluster_kubeconfig.go
```diff
@@ -1,18 +1,25 @@
...
 
 	"github.com/spf13/cobra"
 	"github.com/xetys/hetzner-kube/pkg/clustermanager"
 	"github.com/xetys/hetzner-kube/pkg/hetzner"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+const (
+	defaultContext = "kubernetes-admin@kubernetes"
```
mavimo: I'm not sure on why we are defining this `defaultContext`, maybe context should be "cluster specific", like `kubernetes-admin@{CLUSTER_NAME}`?

cmd/cluster_kubeconfig.go
```diff
@@ -45,41 +53,153 @@ Example 4: hetzner-kube cluster kubeconfig -n my-cluster -p > my-conf.yaml # pri
...
 		} else {
-			fmt.Println("create file")
+			kubeConfigContent = newKubeConfig
+			isSanitized = true
+		}
 
-			usr, _ := user.Current()
-			dir := usr.HomeDir
-			path := fmt.Sprintf("%s/.kube", dir)
+		if merge, _ := cmd.Flags().GetBool("merge"); merge && isSanitized {
```
mavimo: Ignored error, I know that some command ignore error from flag, but I think is not a good practice, WDYT?

cmd/cluster_kubeconfig.go
```diff
@@ -45,41 +53,153 @@ Example 4: hetzner-kube cluster kubeconfig -n my-cluster -p > my-conf.yaml # pri
...
-			if _, err := os.Stat(kubeconfigPath); !force && err == nil {
-				fmt.Println("There already exists a kubeconfig. Overwrite? (use -f to suppress this question) [yN]:")
-				r := bufio.NewReader(os.Stdin)
-				answer, err := r.ReadString('\n')
-				FatalOnError(err)
-				if !strings.ContainsAny(answer, "yY") {
-					log.Fatalln("aborted")
-				}
-			}
+		if save, _ := cmd.Flags().GetBool("save"); save {
```
mavimo: Ignored error

cmd/cluster_kubeconfig.go
```diff
@@ -45,41 +53,153 @@ Example 4: hetzner-kube cluster kubeconfig -n my-cluster -p > my-conf.yaml # pri
...
 
-			ioutil.WriteFile(kubeconfigPath, []byte(kubeConfigContent), 0755)
+			targetPath := fmt.Sprintf("%s/.kube/%s.yaml", GetHome(), provider.GetCluster().Name)
+			if target, _ := cmd.Flags().GetString("target"); target != "" {
```
mavimo: Ignored error

cmd/cluster_kubeconfig.go
```diff
@@ -45,41 +53,153 @@ Example 4: hetzner-kube cluster kubeconfig -n my-cluster -p > my-conf.yaml # pri
...
-			fmt.Println("kubeconfig configured")
+			return
 		}
+
+		fmt.Println(kubeConfigContent)
 	},
 }
 
+// Write kubeConfig to destination
+func doConfigWrite(dst string, kubeConfig string) (err error) {
```
mavimo:
```suggestion
func doConfigWrite(dst string, kubeConfig string) error {
```

cmd/cluster_kubeconfig.go
```diff
@@ -45,41 +53,153 @@ Example 4: hetzner-kube cluster kubeconfig -n my-cluster -p > my-conf.yaml # pri
...
+
+	if _, err := os.Stat(path.Dir(dst)); os.IsNotExist(err) {
+		os.MkdirAll(path.Dir(dst), 0755)
+	}
+	return ioutil.WriteFile(dst, []byte(kubeConfig), 0755)
+}
+
+// Create backup of current kubeCongig
+func doConfigCopyBackUp(src string) (err error) {
```
mavimo:
```suggestion
func doConfigCopyBackUp(src string) error {
```

cmd/util.go
md2k: Yeah, this is debug tools, so can be removed before merge. :)

cmd/cluster_kubeconfig.go
md2k: this is not what we going to set, this is what we looking to rename. regardless cluster, this context set by default during kubebootstrap, so all our configs with this context. maybe i should change name to `default_pattern` then.

md2k:
we can cover it with tests for sure, only i'm not sure it is truly necessary at this part (but it is my vision).
To make more complex logic..., can you describe a bit in depth what you have in mind about it?
Because i do not want abuse tool with full config management.

My idea to cover next scenario:
1) i'm already have several clusters i'm working on in my kubeconfig
2) i'm create new cluster
3) i want this new cluster be available in my config, so i do a merge
3.1) create backup before merge (since the timestamped with seconds, each run will produce new backup, so user safe because he have history of configs)
3.2) merge my sanitized config into main config (config by this tool always created with prefix `hetzner`-ClusterName)
I think in this case it hard to break something, because merge logic completely done with native library and follow next steps:
1) load in 1 or more separate configs to Config structure
2) if config file you provided empty, not exists or broken - ignore it
3) if config file have same context,cluster,authinfo name merge them in order they loaded
For example: you merging `config1` `config2` `config3`
if config2 have same context as in config1, information of `config1` will be overridden by `config2`. 
if apply it to `hetzner_kube` tool, we can update only what `hetnzer_kube`knows, and only if you specify merge flag. if you want to merge , probably you know what you are doing.
One thing only we need to add probably when we destroy cluster also delete it from kubeconfig (but this need to be as additional flag for `cluster delete` i think).

Another way, is implement basic functionality to manage config by this tool.
But then, `kubeconfig` command should be moved from `cluster -> kubeconfig` to standalone and implements next commands:
```
kubeconfig -|
                      |- save <ClusterName/ContextName> --config(-c) /path/to/save
                      |- delete <ClusterName/ContextName> --config(-c) /path/to/save
                      |- merge <ClusterName/ContextName> --config(-c) /path/to/save
                      |- list --config(-c) /path/to/save
```
so we can manage kubeconfig(s), but i really do not want do that, because this is out of scope of this tool and all this functionality already built-in to `kubectl` and for more complex merge management there is plugin for `kubectl`

md2k:
Other suggestions i'll fix today evening.

cmd/cluster_kubeconfig.go
md2k: this done to not assign error directly with `:=`
for example if you going to use:
```
if source, err := os.Open(src); err != nil {
		return
}
```
`source` will available only inside condition, and we want have access to it outside of condition.
if `source` declared, then use of `:=` will rise error.
if to remove declaration of `err` from function then i need also use condition differently as:
```
source, err := os.Open(src)
if err != nil {
   return
}
```
While both variants completely good, first variant probably looks bit nicer.
But here this is how i do usually. also no need do `return` with exact parameters :)
If still want me to use your suggestion, let me know

mavimo: JM2C: the first variant is a bit concise, but make code more hard to understand. We can declare `err` just before the usage (as we do for `source`) and return the error (eventually decorated with extra info), like:
```go
var source, destination *os.File
var err error
if source, err = os.Open(src); err != nil {
    return fmt.Errorf("unable to get source file: %v", err)
}
```
or use the second proposal:
```go
source, err := os.Open(src)
if err != nil {
    return fmt.Errorf("unable to get source file: %v", err)
}
```
IMHO the second variant is more comprehensible when someone look the code for the first time (new contributors, ppl that try to investigate an issue, the user that write the code after few weeks that write the code..).
Off course is my opinion, maybe a 3rd opinion should be useful ;) cc/ @xetys

xetys: I like the second more, as I'm not a fan of C style declaring variables in the header and than doing the code later. The second one is far better to read and understand, and an actual goal of go to make it looking like this

md2k:
heh, weekend was busy. i'll try to finalize today-tomorrow this PR

cmd/cluster_kubeconfig.go
md2k: it depends, as i said, all variants is valid in go. but i can agree , i'm always judging from my 5+ experience in go, and for me personally read some-one code in this scope absolutely transparent. but if majority vote for
```
if err != nil {
    return err
}
```
not a problem :D

md2k: @mavimo @xetys check my comment later below, maybe we really need change way kubeconfig command work, instead of flag usage to print/save/merge , better to change it to sub-command and do `kubeconfig save|print|merge` and then each of subcommand will have own options if necessary?

md2k:
found few bugs in current merge logic, so yeah, need to do more complex logic i think. by some reason it not always merge it in right direction. going to dive into a bit more

md2k:
nvm here, just a rebase from master and fix for conflicts.

cmd/cluster_kubeconfig.go
xetys: If you are sure you properly defined the flag, you can easily ignore this error. I did this several times. The error which might occur here is if the flag itself doesn't exist

cmd/cluster_kubeconfig.go
xetys: same here, if they are defined, no error should occur

xetys:
Hey,

I'm up to release 0.5 soon and would like to merge this PR.

I see the following two ways:

* simple: just fix the merge conflict and wait for @mavimo if he is okay with the current state
* advanced: I like the proposal with the sub-command structure, as I would love to add features like user management later (you already can use `kubeadm alpha kubeconfig user --client-name=...` to add new users to the cluster, which would make sense to pass to the kubeconfig sub-command, too)

Decide which is better for you, how much time you can spend on this right now.

xetys:
hey, what's up here? I would love to see this PR ready for merge

Issue #267, opened 2019-03-30
etcd management


xetys:
This issue just tracks a work in progress.

It is useful for HA clusters (both, level3 and level4) to have a backup and restore utility for etcd.

In basic:

``` bash
hetzner-kube cluster etcd backup my-cluster # will create a snapshot 2019-03-29-12-00.db
hetzner-kube cluster etcd backup my-cluster -s my-snapshot # will create a snapshot my-snapshot.db
hetzner-kube cluster etcd restore my-cluster -s my-snapshot # will restore a snapshot of etcd to that state
```

This is even more useful, if you copy snapshots from other clusters to perform a migration or test stuff.

Will arrive as PR the next hours

Issue #266, opened 2019-03-25
Unmet dependencies on cluster create (kubernetes-cni (= 0.6.0) but 0.7.5-00 is to be installed)


cwoebker:
I just tried setting up a Kubernetes cluster with Hetzner Cloud.
I ran into a similar issue as #195.

```
run failed
command:apt-get install -y docker-ce kubelet=1.13.4-00 kubeadm=1.13.4-00 kubectl=1.13.4-00 kubernetes-cni wireguard linux-headers-$(uname -r) linux-headers-virtual
stdout:Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 kubeadm : Depends: kubernetes-cni (= 0.6.0) but 0.7.5-00 is to be installed
 kubelet : Depends: kubernetes-cni (= 0.6.0) but 0.7.5-00 is to be installed
E: Unable to correct problems, you have held broken packages.

err:Process exited with status 100
```

Possible related to: https://github.com/kubernetes/kubernetes/issues/75683 &  https://github.com/kubernetes/kubernetes/issues/75701

hyperbohlic:
@cwoebker Had the same issue. Quick fix is to change kubernetes-cni to kubernetes-cni=0.6.0-00 in /pkg/clustermanager/provision_node.go:238

gedaiu:
I have the same issue here.

andreymaznyak:
Any updates?

eaydogan:
I fixed this issue on my forked repo.
If anyone interest you can get fix that repo.

https://github.com/eaydogan/hetzner-kube

Here is the commit diff:

https://github.com/eaydogan/hetzner-kube/commit/9e688d7283d1c3f209a4a7c1a0ec8b6e7211aba9

<img width="934" alt="Screen Shot 2019-04-29 at 14 38 18" src="https://user-images.githubusercontent.com/1244956/56903684-6fd6b180-6aa5-11e9-8d39-570c6b574163.png">


<img width="1159" alt="Screen Shot 2019-04-29 at 17 21 25" src="https://user-images.githubusercontent.com/1244956/56903709-7e24cd80-6aa5-11e9-86f0-ac27fad11bcb.png">

[...]

Issue #185, opened 2018-09-07
Kubernetes API Server count flag is statically set to 3


galexrt:
As we talked today here is the issue 😄 

https://github.com/xetys/hetzner-kube/blob/add8676eb2fab01bad7022400b4a1de762bc8f2b/pkg/clustermanager/cluster.go#L386-L392

I have an old patch ready which I'm happy to update to latest and create a PR for.

https://github.com/galexrt/hetzner-kube/blob/0211b58e1ace3f59e0dc26d001211f17b7db2f38/pkg/clustermanager/cluster.go#L424-L431

PR #268, opened 2019-03-30
etcd management, phases & fixes

xetys:
a multi PR with the following contents:

quickfix for #266 
introduction of cluster etcd management: #267 
fixing static apiserver-count #185 
introducing phases system #269


mavimo:
@xetys this PR move the project one level up, nice!!! I'll take some time for me to review it (for sure not before the next weekend)...

cmd/cluster.go
```diff
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"errors"
+	"fmt"
```
mavimo: Add empty line between std lib package and external
```suggestion
	"fmt"

```

cmd/cluster.go
```diff
@@ -20,3 +22,20 @@ are hopefully coming soon.'`,
 func init() {
 	rootCmd.AddCommand(clusterCmd)
 }
+
+// validateClusterInArgumentExists checks if a cluster name is present in arguments and the cluster can be found
+func validateClusterInArgumentExists(cmd *cobra.Command, args []string) error {
+
+	name := args[0]
+
+	if name == "" {
```
mavimo: This should be omitted if we force cobra command to 1 positional argument using `Args` configuration, like:
```go
var clusterCmd = &cobra.Command{
	Use:   "cluster",
	// ...
	Args: cobra.ExactArgs(1),
	// ...
}
```
We should also use a more sophisticated validation adding `ValidArgs`, that report the list of available cluster we have, eg:
```go
var clusterCmd = &cobra.Command{
	Use:   "cluster",
	// ...
	ValidArgs: AppConf.Config.AllClustersName(),
	Args: cobra.ExactValidArgs(1),
	// ...
}
```

cmd/cluster_create.go
```diff
@@ -3,6 +3,7 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"github.com/xetys/hetzner-kube/cmd/phases"
```
mavimo: move after stdlib packages

cmd/cluster_etcd_backup.go
```diff
@@ -0,0 +1,37 @@
...
+	Short:   "creates a backup of the etcd cluster. If no name is provided, a current datetime string is used",
+	PreRunE: validateClusterInArgumentExists,
+	Run: func(cmd *cobra.Command, args []string) {
+		snapshotName, _ := cmd.Flags().GetString("snapshot-name")
+		etcdManager := getEtcdManager(cmd, args)
+
+		err := etcdManager.CreateSnapshot(snapshotName)
+
+		if err != nil {
+			fmt.Println(err)
```
mavimo: Also return error code if we have error and we trap it, otherwise return error to cobra. RunE to allow the cobra library to return formatted message and error code, eg:

```go
var backupCmd = &cobra.Command{
	// ...
	RunE: func(cmd *cobra.Command, args []string) error {
		snapshotName, _ := cmd.Flags().GetString("snapshot-name")
		etcdManager := getEtcdManager(cmd, args)

 		return etcdManager.CreateSnapshot(snapshotName)
	}
```

cmd/cluster_etcd_restore.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+var restoreCmd = &cobra.Command{
+	Use:     "restore",
+	Short:   "restores an etcd snapshot",
+	PreRunE: validateEtcdRestoreCmd,
+	Run: func(cmd *cobra.Command, args []string) {
```
mavimo: Use `RunE` and return error on failure

cmd/cluster_phase_install_masters.go
```diff
@@ -0,0 +1,67 @@
...
+		_, cluster := AppConf.Config.FindClusterByName(clusterName)
+		provider := hetzner.NewHetznerProvider(AppConf.Context, AppConf.Client, *cluster, AppConf.CurrentContext.Token)
+		masterNode, err := provider.GetMasterNode()
+		FatalOnError(err)
+		err = AppConf.SSHClient.(*clustermanager.SSHCommunicator).CapturePassphrase(masterNode.SSHKeyName)
+		FatalOnError(err)
+		coordinator := pkg.NewProgressCoordinator()
+
+		for _, node := range provider.GetAllNodes() {
+			steps := 3
```
mavimo: IMHO we make the code more comprehensible if we do no use code like `step := 3` that hide what we are supposing to do here. Maybe we can iterate over step, add a step name and check name here? Or we can use a chained list on step definition so we can know what's the next step?

cmd/cluster_phase_etcd.go
```diff
@@ -0,0 +1,29 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/xetys/hetzner-kube/cmd/phases"
+	"github.com/xetys/hetzner-kube/pkg"
+)
+
+func init() {
+	command := declarePhaseCommand("etcd", "setups a etcd cluster", func(cmd *cobra.Command, args []string) {
```
mavimo: I know that is a good idea be DRY but IMHO we should declare the command without helper like `declarePhaseCommand` and moving function outside init (imho should be awesome if we are able to move all this stuff in a package that can be invoked from the cmd part of the application)

cmd/cluster_phase_etcd.go
```diff
@@ -0,0 +1,29 @@
...
+		provider, clusterManager, coordinator := getCommonPhaseDependencies(6, cmd, args)
+		keepData, _ := cmd.Flags().GetBool("keep-data")
+
+		phase := phases.NewEtcdSetupPhase(clusterManager, provider, phases.EtcdSetupPhaseOptions{KeepData: keepData})
+
+		if phase.ShouldRun() {
+			err := phase.Run()
+			FatalOnError(err)
```
mavimo: Maybe I'm a bit ripetitive but using RunE and return error when we should stop command is a better approach than use custom function like `FatalOnError` ;)

cmd/cluster_phase_etcd.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/xetys/hetzner-kube/cmd/phases"
+	"github.com/xetys/hetzner-kube/pkg"
+)
+
+func init() {
+	command := declarePhaseCommand("etcd", "setups a etcd cluster", func(cmd *cobra.Command, args []string) {
+		provider, clusterManager, coordinator := getCommonPhaseDependencies(6, cmd, args)
```
mavimo: Using 6 here is really confusing, I suggest to move into a const list (maybe iota?) to define a more comprehensible name using a liked list

mavimo:
@xetys I add some note. Some should be repeated on code but IMHO is not useful until a decision on how to proceed.
There are three main topic IMHO:
1. Use cobra feature to argument validation/command error reporting instead of custom commands
1. Use chained list to define `PhaseChain` so each command is aware of the next one and can invoke the "execute next" operation if needed.
1. Better manage how we should start on command execution when we start from a specific task

I'll like also to discuss some folder structure organisation (eg grouping phases command in a single folder) and package organisation (eg move the actions outside the `cmd` package and move into `pkg` folder, than invoke it on commands, imho separate operational task from the command that define how we expose the UI (cmd should be a kind of ui, but we should expose it also a web page or WS) is better approach.

PS: I REALLY like the work you do on this PR, ad hope that my topic as keep as suggestion that can help to improve the project and not as critique on your work ;)

cmd/cluster_phase_install_masters.go
xetys: The issue with the steps here is, that it's currently a pity with the UIProgress. If you once ran over 100% before it is actually done, it sends a `done` signal to the channel and the bar synchronization is not really working. This leads to the progress always showing some work, as it actually is already complete. You must know the steps before.

I'm also considering a different way of showing progress. Initially, I liked the UI Progress, but it is so ugly in handling. WDYT?

cmd/cluster.go
xetys: That is a good hint. I will try to solve it partially when I can, but actually, this is a little bit out of the scope of this PR. I think replacing all validations (because this file only defines the validation function, but the `cluster` root command doesn't use it. Other commands already do. But yes, we should do this at some point.

xetys:
Well, it took me some time to start. I did some of your hints. However, stuff like redesign the PreRunE to valid args or better chain design is something I would like to see in a different PR.

Issue #259, opened 2019-03-12
CSI Driver support


md2k:
Hi @mavimo  i think your last PR GH-258  for issue #256 not fully fix the problem.
We need add feature-gate options to kube-apiserver as well, otherwise CSI plugin failed with messages i described here (https://github.com/hetznercloud/csi-driver/issues/26)

mavimo:
@md2k I'll try to investigate it since I'm able to setup PVC on a new cluster created with the the code submitted in #258, I'll try to investigate the issue you reported ASAP, thx for feedback!

md2k:
i'll double check again today evening, maybe i'm overlooked something.

md2k:
Hi @mavimo  maybe you can advice what to do with CSI?
```
I0312 18:52:55.121932       1 csi_handler.go:524] Can't get CSINodeInfo k8s-dev-worker-04: csinodeinfos.csi.storage.k8s.io "k8s-dev-worker-04" not found
I0312 18:52:55.121981       1 csi_handler.go:388] Saving attach error to "csi-4f85d0c027dcff14997168efbc3674ba47e378f12476131bbd23449ee74359bc"
I0312 18:52:55.138679       1 csi_handler.go:398] Saved attach error to "csi-4f85d0c027dcff14997168efbc3674ba47e378f12476131bbd23449ee74359bc"
I0312 18:52:55.138740       1 csi_handler.go:103] Error processing "csi-4f85d0c027dcff14997168efbc3674ba47e378f12476131bbd23449ee74359bc": failed to attach: node "k8s-dev-worker-04" has no NodeID annotation
I0312 18:52:55.138671       1 controller.go:139] Ignoring VolumeAttachment "csi-4f85d0c027dcff14997168efbc3674ba47e378f12476131bbd23449ee74359bc" change
I0312 18:55:21.106674       1 controller.go:173] Started VA processing "csi-4f85d0c027dcff14997168efbc3674ba47e378f12476131bbd23449ee74359bc"
I0312 18:55:21.106750       1 csi_handler.go:93] CSIHandler: processing VA "csi-4f85d0c027dcff14997168efbc3674ba47e378f12476131bbd23449ee74359bc"
I0312 18:55:21.106766       1 csi_handler.go:120] Attaching "csi-4f85d0c027dcff14997168efbc3674ba47e378f12476131bbd23449ee74359bc"
[...]

PR #272, opened 2019-04-07
Fix CSI issue and upgrade to kubernetes 1.14.0

mavimo:
This PR will update to kubernetes 1.14.0 and fix issue on CSI.

Actually tested with a new cluster, than create a secret with hetzner token:
```
cat >secret.yaml <<EOF
---
kind: Secret
apiVersion: v1
metadata:
  namespace: kube-system
  name: hcloud-csi
data:
  token: {ADD_TOKEN_BASE64_ENCODED_HERE}
EOF
```
and deploy all the required stuff:
```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/csi-api/release-1.13/pkg/crd/manifests/csidriver.yaml
kubectl apply -f https://raw.githubusercontent.com/kubernetes/csi-api/release-1.13/pkg/crd/manifests/csinodeinfo.yaml
kubectl apply -f https://gist.githubusercontent.com/mavimo/399d9f783a59653bac71c8ddbf886757/raw/bc9ff2bcc6c38fea27f2586ce5da0cfa2a6e004e/hetzner-csi.yaml
```
at the end create a new pod using the PV, eg:
```
cat >test-csi.yaml <<EOF
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: csi-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
  storageClassName: hcloud-volumes
---
kind: Pod
apiVersion: v1
metadata:
  name: my-csi-app
spec:
  containers:
    - name: my-frontend
      image: busybox
      volumeMounts:
      - mountPath: "/data"
        name: my-csi-volume
      command: [ "sleep", "1000000" ]
  volumes:
    - name: my-csi-volume
      persistentVolumeClaim:
        claimName: csi-pvc
EOF;

```

and deploy it:
```
kubectl apply -f test-csi.yaml
```

that's work without issue on logs (I inspected with `journalctl -u kubelet` on master and worker nodes).

### DIFF
--- a/pkg/clustermanager/configs.go
+++ b/pkg/clustermanager/configs.go
@@ -9,11 +9,15 @@ import (
 func GenerateMasterConfiguration(masterNode Node, masterNodes []Node, etcdNodes []Node) string {
 	masterConfigTpl := `apiVersion: kubeadm.k8s.io/v1beta1
 kind: ClusterConfiguration
+kubernetesVersion: stable
 networking:
   serviceSubnet: "10.96.0.0/12"
   podSubnet: "10.244.0.0/16"
   dnsDomain: "cluster.local"
 apiServer:
+  featureGates:
+    CSINodeInfo: true
+    CSIDriverRegistry: true
   certSANs:
     - 127.0.0.1
     - %s
--- a/pkg/clustermanager/configs_test.go
+++ b/pkg/clustermanager/configs_test.go
@@ -9,11 +9,15 @@ import (
 func TestGenerateMasterConfiguration(t *testing.T) {
 	expectedConf := `apiVersion: kubeadm.k8s.io/v1beta1
 kind: ClusterConfiguration
+kubernetesVersion: stable
 networking:
   serviceSubnet: "10.96.0.0/12"
   podSubnet: "10.244.0.0/16"
   dnsDomain: "cluster.local"
 apiServer:
+  featureGates:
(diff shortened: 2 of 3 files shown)


mavimo:
refs https://github.com/xetys/hetzner-kube/issues/259

@md2k @suchwerk can you check and let me know if it work also on your case or you still see any issue on it? Thx!

suchwerk:
Cant test it right now, but looks pretty good.

BitPhinix:
Tested it, works like a charm

pkg/clustermanager/configs.go
```diff
@@ -9,11 +9,15 @@ import (
 func GenerateMasterConfiguration(masterNode Node, masterNodes []Node, etcdNodes []Node) string {
 	masterConfigTpl := `apiVersion: kubeadm.k8s.io/v1beta1
 kind: ClusterConfiguration
+kubernetesVersion: stable
```
xetys: CC is crying on this one, this makes the function quite big. Maybe moving the template contents into a variable, instead of having it inside the function? WDYT?

mavimo: IMHO is not a good solution since we move the test less comprehensible, in this case I'll prefer ignore CC issue, but if you think still useful I'll move this evening.

Issue #294, opened 2019-11-12
Installing "hcloud-controller-manager" fails


drallgood:
```% hetzner-kube cluster addon install --name <name> hcloud-controller-manager
2019/11/12 15:06:09 installing addon hcloud-controller-manager
2019/11/12 15:06:36 run failed
command:kubectl -n kube-system patch ds kube-flannel-ds --type json -p '[{"op":"add","path":"/spec/template/spec/tolerations/-","value":{"key":"node.cloudprovider.kubernetes.io/uninitialized","value":"true","effect":"NoSchedule"}}]'
stdout:Error from server (NotFound): daemonsets.apps "kube-flannel-ds" not found

err:Process exited with status 1
```

And indeed I don't see flannel anywhere.

alexzimmer96:
@drallgood 
Should be fixed with the next release. If you want to try the fix you can build the tool by yourself and install the addon.

xetys:
As the release is out, I close this

PR #295, opened 2019-11-21
Fix hcloud controller manager

alexzimmer96:
This should fix issue #294 

In the current version, the "hcloud-controller-manager"-addon wants to update the (non-existing) flannel Daemonset. This has been corrected to canal.

It also tries to use version 1.2.0 of the hcloud-cloud-controller-manager which is not suitable for Kubernetes 1.6. I've pushed this version.

And because thats not enough, there is also a bug in the new cloud-controller-manager deployment spec which requires a "network"-key inside the hcloud secret even if it is not used inside the deployment. I added a workaround to remove this secret from the deployment-spec using a patch-request.

Tested this on my cluster


alexzimmer96:
@xetys

xetys:
LGTM

PR #296, opened 2019-11-21
randomize the order of datacenters

drallgood:
When adding new (worker) nodes, the nodes are always created in the same data centers based on the provided order (default [fsn1-dc8,nbg1-dc3,hel1-dc2,fsn1-dc14]).
And while that works fine upon initial cluster creation, evenly distributing nodes across data centers it does not work when creating nodes after the fact. They're always created in the same data center.

This PR fixes that by randomizing the provided data center list while still making sure the nodes are evenly distributed.

### DIFF
--- a/pkg/hetzner/hetzner_provider.go
+++ b/pkg/hetzner/hetzner_provider.go
@@ -6,9 +6,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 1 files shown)


xetys:
LGTM, I will merge it if you fix the CC issue

drallgood:
Done @xetys

PR #297, opened 2019-11-21
adding cloud-init script to add-worker cmd

drallgood:
This allows a custom user-init scripts to be used when adding new nodes, e.g. to set up floating IPs or things like ansible.

I mostly tried copying what was already there. Suggestions are always very welcome.

### DIFF
--- a/cmd/cluster_add_worker.go
+++ b/cmd/cluster_add_worker.go
@@ -4,10 +4,10 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
 	"strconv"
 	"strings"
 	"time"
-	"os"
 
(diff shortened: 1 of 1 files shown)


xetys:
hey,

what is the use-case here? You create a cluster without cloud-init, but then add a cloud-init for a worker, do I get this right? Did you try if that works?

drallgood:
@xetys
Exactly. Or you want a different config than before for these workers. 

Yes, it works. used it twice yesterday

drallgood:
@xetys Is that something you want to incorporate?

xetys:
thanks for reminding me...

LGTM

PR #308, opened 2019-12-26
Make CNI configurable and add Calico as an option

MattiasGees:
This makes CNI configurable so people can start adding other CNI support to it. Calico is also now one of the options to use. 
The default is still Canal and it will be picked automatically when no option is given.

### DIFF
--- a/pkg/clustermanager/cluster.go
+++ b/pkg/clustermanager/cluster.go
@@ -15,6 +15,7 @@ type Manager struct {
 	nodes            []Node
 	clusterName      string
 	cloudInitFile    string
+	CNI              string
 	eventService     EventService
 	nodeCommunicator NodeCommunicator
 	clusterProvider  ClusterProvider
@@ -36,7 +37,7 @@ const (
 )
 
 // NewClusterManager create a new manager for the cluster
-func NewClusterManager(provider ClusterProvider, nodeCommunicator NodeCommunicator, eventService EventService, name string, haEnabled bool, isolatedEtcd bool, cloudInitFile string) *Manager {
+func NewClusterManager(provider ClusterProvider, nodeCommunicator NodeCommunicator, eventService EventService, name string, haEnabled bool, isolatedEtcd bool, cloudInitFile string, cni string) *Manager {
 	manager := &Manager{
 		clusterName:      name,
 		haEnabled:        haEnabled,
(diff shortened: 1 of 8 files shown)


pkg/clustermanager/cluster.go
```diff
@@ -15,6 +15,7 @@ type Manager struct {
 	nodes            []Node
 	clusterName      string
 	cloudInitFile    string
+	CNI              string
```
xetys: In GoLang upper case capped fields are public, while lower case capped fields are private. In this type, we use mostly private fields, so `CNI` should be `cni`.

pkg/clustermanager/cluster.go
```diff
@@ -45,6 +46,7 @@ func NewClusterManager(provider ClusterProvider, nodeCommunicator NodeCommunicat
 		eventService:     eventService,
 		nodeCommunicator: nodeCommunicator,
 		clusterProvider:  provider,
+		CNI:              cni,
```
xetys: same here

pkg/clustermanager/cluster.go
```diff
@@ -74,6 +76,7 @@ func (manager *Manager) Cluster() Cluster {
 		IsolatedEtcd:  manager.isolatedEtcd,
 		CloudInitFile: manager.cloudInitFile,
 		NodeCIDR:      manager.clusterProvider.GetNodeCidr(),
+		CNI:           manager.CNI,
```
xetys: same here

pkg/clustermanager/cluster.go
MattiasGees: Is fixed now

pkg/clustermanager/cluster.go
MattiasGees: Is fixed now

pkg/clustermanager/cluster.go
MattiasGees: Is fixed now

xetys:
Do you think you can fix the CC issues, too?

MattiasGees:
I have been looking at how to optimise it and not sure how to do it. Pointers are welcome :)

`validateClusterCreateFlags` only adds 3 lines. Even when I move it to another method it will stay the same

`InstallMasters` adds 5 lines, but I am also not sure it is worth moving that part to another method

MattiasGees:
@xetys wdyt?

xetys:
I would say ignore CC, but pls resolve the merge conflicts

MattiasGees:
Merge conflicts have been resolved.

PR #310, opened 2019-12-28
Make sure Kubernetes version is the same

### LINKED ISSUE #306: Getting 2 different Kubernetes versions
When doing an install from the current master branch, it mixes 2 Kubernetes versions.

When doing `kubectl version` I am getting the following:

```
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.4", GitCommit:"224be7bdce5a9dd0c2fd0d46b83865648e2fe0ba", GitTreeState:"clean", BuildDate:"2019-12-11T12:37:43Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

But when doing `kubectl get nodes` I am getting this
```
NAME                   STATUS   ROLES    AGE   VERSION
kubernetes-master-01   Ready    master   35m   v1.16.2
kubernetes-worker-01   Ready    <none>   35m   v1.16.2
kubernetes-worker-02   Ready    <none>   35m   v1.16.2
```

I checked the kubeadm and kubectl package on the master server and they are both `1.16.2-00`. I think it has to do with `kubernetesVersion: stable` in `/root/master-config.yaml` when doing the `kubeadm init`. Should we fix the version in there?

MattiasGees:
This makes sure that the Kubernetes version is the same everywhere. This is a quick POC at the moment.

@xetys what do you think about this idea? Should we also make the Kubernetes version configurable by a CLI flag?

This fixes #306

### DIFF
--- a/pkg/clustermanager/provision_node.go
+++ b/pkg/clustermanager/provision_node.go
@@ -1,7 +1,6 @@
 package clustermanager
 
 import (
-	"flag"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


pkg/clustermanager/provision_node.go
```diff
@@ -10,24 +9,23 @@ import (
...
+	clusterName       string
+	node              Node
+	communicator      NodeCommunicator
+	eventService      EventService
+	kubernetesVersion string
 }
 
 // NewNodeProvisioner creates a NodeProvisioner instance
-func NewNodeProvisioner(clusterName string, node Node, communicator NodeCommunicator, eventService EventService) *NodeProvisioner {
+func NewNodeProvisioner(clusterName string, node Node, communicator NodeCommunicator, eventService EventService, kubernetesVersion string) *NodeProvisioner {
```
MattiasGees: Maybe we should give the manager object as argument here?

xetys: How would this look like?

MattiasGees: I made a commit to do this

PR #315, opened 2020-02-01
Fix file permissions for kube-hetzner

### LINKED ISSUE #188: config with API tokens is world-readable by default
The config file should be created with permissions 0600, so that other users of the machine cannot read the secret API keys.

MattiasGees:
This fixes #188


xetys:
LGTM

PR #316, opened 2020-02-17
Bump hcloud-cloud-controller-manager version

reixd:
The old v1.5.0 manifest is no longer available, therefore need to change the version to the current release.

### DIFF
--- a/pkg/addons/addon_hcloud_controller_manager.go
+++ b/pkg/addons/addon_hcloud_controller_manager.go
@@ -70,7 +70,7 @@ Environment="KUBELET_EXTRA_ARGS=--cloud-provider=external"
 	FatalOnError(err)
 	_, err = addon.communicator.RunCmd(*addon.masterNode, fmt.Sprintf("kubectl -n kube-system create secret generic hcloud --from-literal=token=%s", addon.provider.Token()))
 	FatalOnError(err)
-	_, err = addon.communicator.RunCmd(*addon.masterNode, "kubectl apply -f  https://raw.githubusercontent.com/hetznercloud/hcloud-cloud-controller-manager/master/deploy/v1.5.0.yaml")
+	_, err = addon.communicator.RunCmd(*addon.masterNode, "kubectl apply -f  https://raw.githubusercontent.com/hetznercloud/hcloud-cloud-controller-manager/master/deploy/v1.5.1.yaml")
 	FatalOnError(err)
 	// This is needed cause there is a bug inside the hcloud-cloud-controller-manager deployment spec.
 	// The env-variable "network" is not marked as optional but is also not needed for the deployment.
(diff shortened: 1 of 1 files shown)


mavimo:
@reixd thanks! The issue that I see is that we are using the `master` branch, then when master is updated the code will change. Can we refer to the tagged version (eg: https://github.com/hetznercloud/hcloud-cloud-controller-manager/tree/v1.5.0/deploy or https://github.com/hetznercloud/hcloud-cloud-controller-manager/tree/v1.5.1/deploy ) so we are keeping the right reference to the file also if the code on master change?
Thx!!

pkg/addons/addon_hcloud_controller_manager.go
```diff
@@ -70,7 +70,7 @@ Environment="KUBELET_EXTRA_ARGS=--cloud-provider=external"
 	FatalOnError(err)
 	_, err = addon.communicator.RunCmd(*addon.masterNode, fmt.Sprintf("kubectl -n kube-system create secret generic hcloud --from-literal=token=%s", addon.provider.Token()))
 	FatalOnError(err)
-	_, err = addon.communicator.RunCmd(*addon.masterNode, "kubectl apply -f  https://raw.githubusercontent.com/hetznercloud/hcloud-cloud-controller-manager/master/deploy/v1.5.0.yaml")
+	_, err = addon.communicator.RunCmd(*addon.masterNode, "kubectl apply -f  https://raw.githubusercontent.com/hetznercloud/hcloud-cloud-controller-manager/master/deploy/v1.5.1.yaml")
```
mavimo: refer to `/tree/v1.5.1/` instead of master

pkg/addons/addon_hcloud_controller_manager.go
```diff
@@ -82,7 +82,7 @@ Environment="KUBELET_EXTRA_ARGS=--cloud-provider=external"
 
 // Uninstall performs all steps to remove the addon
 func (addon *HCloudControllerManagerAddon) Uninstall() {
-	_, err := addon.communicator.RunCmd(*addon.masterNode, "kubectl delete -f  https://raw.githubusercontent.com/hetznercloud/hcloud-cloud-controller-manager/master/deploy/v1.5.0.yaml")
+	_, err := addon.communicator.RunCmd(*addon.masterNode, "kubectl delete -f  https://raw.githubusercontent.com/hetznercloud/hcloud-cloud-controller-manager/master/deploy/v1.5.1.yaml")
```
mavimo: refer to `/tree/v1.5.1/` instead of master

pkg/addons/addon_hcloud_controller_manager.go
reixd: wilco. Is also the better solution

mavimo:
@reixd thx for the fix!

@xetys LGTM

PR #317, opened 2020-02-17
Fix bash syntax error

reixd:
We need to apply the storageclass manifest even if the storageclass was not successfully deleted (e.g. because it does not exists)


mavimo:
@reixd thx for your contribution! 💪 
@xetys LGTM

PR #330, opened 2020-06-25
fix external worker flow

xetys:
I fixed some issues when using current hetzner-kube with the external worker mode


pkg/clustermanager/cluster.go
```diff
@@ -373,7 +373,7 @@ func (manager *Manager) InstallWorkers(nodes []Node) error {
 				manager.eventService.AddEvent(node.Name, "registering node")
 				_, err := manager.nodeCommunicator.RunCmd(
 					node,
-					"for i in ip_vs ip_vs_rr ip_vs_wrr ip_vs_sh nf_conntrack_ipv4; do modprobe $i; done"+
```
xetys, line 382: This line wasn't working on a bare-metal node for some reason. Anyone knows why we need this?

Issue #329, opened 2020-06-12
adding workers fails on Wireguard configuration


abarthol:
When adding a worker the install process stops at Wireguard configuration.

`command:systemctl enable wg-quick@wg0 && systemctl restart wg-quick@wg0 && systemctl enable overlay-route.service && systemctl restart overlay-route.service
stdout:Created symlink /etc/systemd/system/multi-user.target.wants/wg-quick@wg0.service → /lib/systemd/system/wg-quick@.service.
Job for wg-quick@wg0.service failed because the control process exited with error code.
See "systemctl status wg-quick@wg0.service" and "journalctl -xe" for details.
`

drallgood:
yep. Ran into the same problem.

Ubuntu borked the wireguard module

Solution is apparently to install hwe (-> newer Kernel):
https://wiki.ubuntu.com/Kernel/LTSEnablementStack

It works, but... don't know how to do that for a node

drallgood:
So.. easiest fix: patch hetzner-kube to use Ubuntu 20.04LTS
I actually upgraded a few of my existing nodes as well that broke (actually the reason why I wanted to recreate one in the first place)

abarthol:
I've tried to use Ubuntu 20.04LTS but I get another error:

`command:for i in ip_vs ip_vs_rr ip_vs_wrr ip_vs_sh nf_conntrack_ipv4; do modprobe $i; done && kubeadm reset -f && kubeadm join 10.0.1.1:6443 --token q5nor9.i7r02nwpwgl1cimm     --discovery-token-ca-cert-hash sha256:4e2d467803467a8aab9c484fa24b0c45db0c865a68c528fdd985b56879afa6c9 

stdout:modprobe: FATAL: Module nf_conntrack_ipv4 not found in directory /lib/modules/5.4.0-28-generic
`

drallgood:
Strange.... I just installed multiple new nodes and it worked just fine...
Anyhow you need the corresponding kernel module

tmemenga:
@abarthol have a look at https://github.com/tmemenga/hetzner-kube/tree/ubuntu-20-04, i was able to get past that error. But i still need to check if the cluster is really operational.

xetys:
If this works, I'd love to see a PR if you don't mind

abarthol:
[...]

PR #336, opened 2020-07-19
Ubuntu 20 04

tmemenga:


tmemenga:
see https://github.com/xetys/hetzner-kube/issues/329

tmemenga:
does not work anymore, we switched to https://github.com/xetys/hetzner-kube/pull/341

